PR #666, opened 2017-02-14
fix issue with new version of tastypie

eshellman:


eshellman:
bug triggered by @egardner. has been there since our move to dj18.

rdhyee:
The code looks good. Let me do merge and see effect on just.

rdhyee:
pushed to production

PR #673, opened 2017-03-21
build instances with trusty

rdhyee:
Moving to `trusty` (14.04 LTS) instead of `precise`


eshellman:
looks superficially ok, anything I should look at closely?

rdhyee:
No need to do anything, @eshellman -- I'm going to merge this PR and build just.  If that goes well, then I'll build production.

PR #696, opened 2017-08-25
remove unneeded requirements

eshellman:


eshellman:
@rdhyee I'm removing some packages that don't seem to be used. Is there anything on the list that you think I ought to put back?

Issue #1096, opened 2026-03-03
OAI harvester not picking up new DOAB books: no cron trigger + null edition errors


rdhyee:
## Summary

The DOAB OAI harvester (`load_doab` management command) has no scheduled trigger and last ran 8 days ago (2026-02-22). It is not blocked by robot rules. The main gaps are: (1) no cron to run it automatically, and (2) a class of DOAB records silently produces `null edition` errors and is never loaded.

---

## Investigation (2026-03-03)

**Prompted by**: Eric's observation that the harvester isn't picking up new DOAB books, and suspicion of unintentional robot limits.

### Finding 1: No scheduled trigger — runs manually only

`crontab -l` (root + ubuntu), `/etc/cron.d/`, and Celery beat all checked — no entry for `load_doab` or `harvest`. The last run was **2026-02-22 23:03–23:10 UTC** per `/var/log/regluit/unglue.it.log`. It hasn't run in 8 days simply because nothing triggered it.

### Finding 2: DOAB OAI endpoint is reachable from prod

```
GET https://directory.doabooks.org/oai/request?verb=Identify → HTTP 200
```

No network, DNS, or TLS issue. The `oaipmh` Python library sends `User-Agent: pyoai`, which is not listed in DOAB's robots.txt block list. DOAB's robots.txt disallows `/discover`, `/browse`, `/statistics` etc. — not `/oai/request`.

### Finding 3: Our bot-blocking cannot affect outbound harvest calls

`robots.txt` and `bot-block.conf` govern **inbound** crawlers. The harvester calls *out* to DOAB's server — our rules have zero effect on it in either direction.

`doab-check.ebookfoundation.org` (146.190.206.51) appears nowhere in Apache logs and is not in iptables — it is not being blocked by unglue.it.

### Finding 4: "null edition" errors — DOAB records with no download URL

The Feb 22 run logged 20+ errors:

```
ERROR regluit.core.loaders.doab[load_doab_oai]: null edition for doab #20.500.12854/171516
ERROR regluit.core.loaders.doab[load_doab_oai]: null edition for doab #20.500.12854/171505
... (20+ more)
```

In `core/loaders/doab.py`, `add_by_doab()` returns `None` when the record's `identifier` list yields no downloadable URLs (the `for dl_url in urls` loop never executes). These are likely DOAB records with chapter-only or reader-only access — no direct ebook download link. They are silently skipped.

### Finding 5: Hardcoded limit of 100 records per run

[...]

PR #1107, opened 2026-03-03
Add timeouts to ContentTyper, load_ebookfile, and get_soup HTTP calls

### LINKED ISSUE #1103: Add HTTP timeouts to ContentTyper / type_for_url in core/models/loader.py
## Problem

PR #1099 added `timeout=(5, 15)` to cover-fetch HTTP calls in `core/loaders/doab.py` and `doab_utils.py`. But a larger timeout gap remains: `type_for_url()` in `core/models/loader.py` is called for **every download URL** in `add_by_doab()` to determine file format, and it makes `requests.head` (with a fallback `requests.get`) calls with **no timeout**.

During a harvest of 800 records, each with one or more download URLs, this means up to 800+ blocking network calls to external publisher servers — with no upper bound on how long each can take.

This is likely a significant contributor to the multi-hour harvest times observed before PR #1099, and may still cause hangs even after that fix lands.

## Where it happens

`core/models/loader.py` — `ContentTyper.content_type()`:

```python
r = requests.head(url, allow_redirects=True)        # no timeout
# ...
r = requests.get(url)                               # no timeout (405 fallback)
r = requests.get(url, verify=False)                 # no timeout (SSL fallback)
```

## Fix

Add `timeout=(5, 15)` to all three `requests` calls in `ContentTyper.content_type()`.

## Related

- Part of #1096 (DOAB harvester investigation)
- Companion to PR #1099 (which fixed cover-fetch timeouts)
- Identified by Gemini and Codex review

rdhyee:
## Summary

Adds `timeout=(5, 15)` to all four `requests` calls in `ContentTyper.content_type()` in `core/models/loader.py`.

## Why

`type_for_url()` is called once per download URL during the DOAB harvest (`add_by_doab()` line ~448) to determine the file format before loading. It makes `requests.head` (with fallback `requests.get`) calls with no timeout. During a full harvest (800+ records), this is potentially 800+ blocking calls to external publisher servers — any one of which can stall indefinitely on an unresponsive host.

This is the same class of problem fixed for cover fetches in PR #1099, but on a higher-frequency code path (called per record, not just for cover images).

## Calls patched

| Location | Call |
|----------|------|
| `handle_ude()` | `requests.get(..., allow_redirects=True)` |
| main path | `requests.head(..., allow_redirects=True)` |
| 405 fallback | `requests.get(url)` |
| SSL fallback | `requests.get(url, verify=False)` |

All use `timeout=(5, 15)`: 5s to establish TCP connection, 15s to receive response.

## Test plan

- [ ] Run `load_doab` on test.unglue.it; confirm no indefinite hangs on unresponsive servers
- [ ] Confirm format detection still works correctly for responsive servers (PDF, EPUB, MOBI)

## Related

- Closes #1103
- Companion to PR #1099 (cover-fetch timeouts)
- Part of #1096 (DOAB harvester investigation)

### DIFF
--- a/core/loaders/soup.py
+++ b/core/loaders/soup.py
@@ -10,12 +10,15 @@ logger = logging.getLogger(__name__)
 def get_soup(url, user_agent=settings.USER_AGENT, follow_redirects=False, verify=True):
     try:
         response = requests.get(url, headers={"User-Agent": user_agent},
-                                allow_redirects=follow_redirects, verify=verify)
+                                allow_redirects=follow_redirects, verify=verify, timeout=(10, 30))
     except requests.exceptions.MissingSchema:
-        response = requests.get('http://%s' % url, headers={"User-Agent": user_agent})
+        response = requests.get('http://%s' % url, headers={"User-Agent": user_agent}, timeout=(10, 30))
     except requests.exceptions.ConnectionError as e:
         logger.error("Connection refused for %s", url)
         logger.error(e)
+        return None
+    except requests.exceptions.Timeout:
+        logger.error("Request timed out for %s", url)
         return None
     if response.status_code == 200:
         soup = BeautifulSoup(response.content, 'lxml')
--- a/core/models/loader.py
+++ b/core/models/loader.py
@@ -85,16 +85,16 @@ class ContentTyper(object):
         def handle_ude(url, ude):
             url = requote(url)
             try:
-                return requests.get(url, allow_redirects=True)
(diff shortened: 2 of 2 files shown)


rdhyee:
**Updated**: ContentTyper read timeout changed from 15s to 60s per Eric's feedback — `timeout=(5, 60)` on all ContentTyper calls. load_ebookfile and get_soup timeouts unchanged (`(10, 60)` and `(10, 30)` respectively).

PR #1123, opened 2026-04-08
Upgrade Django 1.11 → 4.2 LTS

rdhyee:
**Fixes #1110** — Upgrades Django 1.11 → 4.2 LTS across 101 files (+716 −295).

## What this does

Completes the full LTS hop path (1.11 → 2.2 → 3.2 → 4.2) in a single branch. The changes are interdependent — URL routing, middleware, template tags, ORM, auth, and dependency versions all had to move together.

**Key changes:**
- Replace all deprecated Django APIs (`url()`, `ugettext_lazy`, `NullBooleanField`, `{% ifequal %}`, etc.)
- Upgrade `django-registration` 2.x → 3.x (module rename `registration` → `django_registration`)
- Replace `django-jsonfield` with native `models.JSONField`
- Re-parent `IPAddressModelField` onto `PositiveIntegerField` (no-op migration — no schema change)
- Fix vendored deps: django-notification and django-email-change (Python 3 / Django 4.2 compat)
- Add `SESSION_SERIALIZER`, password reset URL redirect, `mimetype=` → `content_type=`
- Bump 9 dependencies (see table below)
- Add BDD smoke test suite (28 scenarios, repo-reproducible)

## Review status

Three independent AI reviews completed + iterative triage:

| Reviewer | Findings | Status |
|----------|----------|--------|
| Claude Code | 5 issues (SESSION_SERIALIZER, password redirect, vendored deps, provisioning gaps) | All fixed |
| Codex | 6 issues (vendored dep pins, mimetype=, Stripe validation, Celery health, Transaction.extra, BDD reproducibility) | Code issues fixed; runtime gates defined |
| Gemini | 5 issues (migrations, Stripe, MIDDLEWARE, URL shadowing, templates) | All triaged as false positives with evidence |

See comment thread for full back-and-forth.

## Deployment plan ([v5](https://github.com/Gluejar/regluit/pull/1123#issuecomment-4216273030) — Codex signed off)

| Phase | Action | Status |
|-------|--------|--------|
| **0. Provisioning** | Merge [regluit-provisioning#25](https://github.com/EbookFoundation/regluit-provisioning/pull/25) (production vars: Python 3.12, Celery 5, `git_branch: "main"`) | Ready for review |
| **1. Prep** | Tag production rollback point. Verify `Transaction.extra` on prod MySQL. | Blocked on Phase 0 |
[...]

PR #1126, opened 2026-04-23
Overlay migration: fix django_comments on_delete state drift

rdhyee:
Refs risk assessment follow-up on [regluit#1123](https://github.com/Gluejar/regluit/pull/1123).

## Why

`manage.py makemigrations --check` reports state drift for the `django_comments` app under Django 4.x:

```
Migrations for 'django_comments':
  venv/.../django_comments/migrations/0004_alter_comment_content_type_alter_comment_user.py
    - Alter field content_type on comment
    - Alter field user on comment
```

**Root cause**: `django_comments.Comment` declares explicit `on_delete` values (`CASCADE` on `content_type`, `SET_NULL` on `user`) but the django-contrib-comments package's shipped migration history was generated under Django 1.x, which didn't require `on_delete`. Django 4.x introspection sees the model-vs-migration-state mismatch and wants to add a recording migration. Upstream django-contrib-comments 2.2.0 (Jan 2022) didn't ship this fix.

## What

Override `MIGRATION_MODULES` to point the `django_comments` app at a regluit-local overlay (`regluit/migrations_overlays/django_comments/`). The overlay contains:

- The 3 upstream migrations copied **verbatim** (0001_initial, 0002_update_user_email_field_length, 0003_add_submit_date_index). Names match upstream so existing `django_migrations` table records are preserved.
- A new `0004_alter_comment_content_type_alter_comment_user` that records the explicit `on_delete` values.

## Safety — this is a state-tracker no-op

Verified on dj42.unglue.it with `sqlmigrate`:

```
$ manage.py sqlmigrate django_comments 0004
-- Alter field content_type on comment
-- (no-op)
-- Alter field user on comment
-- (no-op)
```

**No ALTER TABLE runs.** The actual MySQL FOREIGN KEY constraints (currently `ON DELETE RESTRICT`, inherited from Django 1.x table creation) are not modified. Django's ORM-layer `on_delete` logic (driven by the Python model, not the DB constraint) already produces the correct runtime behavior — it deletes children in Python before issuing the parent DELETE — and continues to do so.

Works on both Django 1.11 (current master, current prod) and Django 4.2 (PR #1123). Safe to land before or after #1123.

## Validation

```
$ env DJANGO_SETTINGS_MODULE=regluit.settings.prod ... \
    manage.py makemigrations --check --dry-run
No changes detected  ✓  (was: reports django_comments drift)
```

## Scope

4 new files + 14 lines in `settings/common.py`. Self-contained. Does not depend on PR #1123; does not affect its scope.

## Future

If upstream django-contrib-comments ever ships an equivalent migration with the same name, this overlay can be removed — migration names would match upstream's, and `django_migrations` applied-state records carry forward.

## Test plan

- [x] `makemigrations --check` clean on Django 4.2 (dj42.unglue.it)
- [ ] `makemigrations --check` clean on Django 1.11 (current master) — deferred; no harm landing without, since Django 1.11's check is less strict about on_delete anyway
- [ ] Quick smoke test: `/comments/` URL responds correctly after deploy

### DIFF
--- a/migrations_overlays/django_comments/0001_initial.py
+++ b/migrations_overlays/django_comments/0001_initial.py
@@ -1,0 +1,71 @@
+from django.db import models, migrations
+from django.conf import settings
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('sites', '0001_initial'),
+        migrations.swappable_dependency(settings.AUTH_USER_MODEL),
+        ('contenttypes', '0001_initial'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='Comment',
+            fields=[
+                ('id', models.AutoField(verbose_name='ID', serialize=False, auto_created=True, primary_key=True)),
+                ('object_pk', models.TextField(verbose_name='object ID')),
+                ('user_name', models.CharField(max_length=50, verbose_name="user's name", blank=True)),
+                ('user_email', models.EmailField(max_length=75, verbose_name="user's email address", blank=True)),
+                ('user_url', models.URLField(verbose_name="user's URL", blank=True)),
+                ('comment', models.TextField(max_length=3000, verbose_name='comment')),
+                ('submit_date', models.DateTimeField(default=None, verbose_name='date/time submitted')),
+                ('ip_address', models.GenericIPAddressField(
+                    unpack_ipv4=True, null=True, verbose_name='IP address', blank=True)),
+                ('is_public', models.BooleanField(default=True,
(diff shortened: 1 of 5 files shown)


eshellman:
we should consider just removing comments

rdhyee:
Closing — superseded by today's Option B decision in [#1123 agenda](https://github.com/Gluejar/regluit/pull/1123): we're shipping #1123 with `django_comments` still wired in, and stripping the app entirely in follow-up #1130 on Django 4.2. The `on_delete` overlay isn't needed if the app is going away.

PR #1137, opened 2026-04-29
Fix #1136: 500 in /accounts/register/ when notarobot input isn't an int

### LINKED ISSUE #1136: 500 error on /accounts/register/: ValueError in clean_notarobot when input has trailing characters
## Symptom

Production is emitting `[Django] ERROR (EXTERNAL IP): Internal Server Error: /accounts/register/` with:

```
ValueError at /accounts/register/
invalid literal for int() with base 10: '14.'

File "/opt/regluit/libraryauth/forms.py" in clean_notarobot
  99.   notarobot = int(self.data["notarobot"])
```

POST contained `notarobot = '14.'` (trailing period). Almost certainly an automated submission — UA was `Firefox/150.0` (a version that doesn't exist) — but a fat-fingered legitimate user could trigger the same crash.

## Root cause

`clean_notarobot` reads from `self.data` (the raw POST `QueryDict`) instead of `self.cleaned_data`:

```python
notarobot = forms.IntegerField(...)        # line 81

def clean_notarobot(self):
    notarobot = int(self.data["notarobot"])  # line 99 — bypasses IntegerField.to_python
```

Because `notarobot` is declared as `forms.IntegerField`, Django would normally have already coerced/rejected the value via the field's own `to_python`. By reaching into `self.data` directly, the code skips that validation, so a non-int string raises `ValueError` instead of `ValidationError` and Django returns 500.

## Secondary issue (same method)

```python
tries = self.data.get("tries", -1)
if str(encoded_answer) != tries:
```

The default `-1` is an `int` while a present value is a `str`, so the comparison shape is inconsistent. The default branch always fails the equality check (which happens to be the desired behavior, but for the wrong reason).

## Proposed fix

Catch the `ValueError` from the `int()` call and convert it to a `forms.ValidationError` so a malformed answer renders as a normal form error instead of a 500. Minimal change, preserves existing behavior for well-formed submissions. PR to follow.

## Impact

- 500 emails to admins on every malformed registration POST (currently noisy on prod due to bot probing).
- Legitimate users typing e.g. `"14."` see a generic error page instead of an inline form error.

rdhyee:
Fixes #1136 — stops the registration form from returning a 500 (and emailing admins) when the `notarobot` POST field can't be parsed as an int (e.g. `'14.'` from a bot probe, or a fat-fingered legitimate user).

## Why

`clean_notarobot` reads `self.data["notarobot"]` (raw POST string) and calls `int()` on it directly, bypassing the `IntegerField`'s own `to_python`. Any non-int value raises `ValueError`, which Django doesn't recognize as a form error and surfaces as a 500. Currently noisy on prod due to bot traffic — see #1136 for the full traceback.

## Change

Wrap the `int()` in try/except and raise `forms.ValidationError` with the existing hint message. Minimal diff; preserves behavior for well-formed submissions.

```python
try:
    notarobot = int(self.data["notarobot"])
except (ValueError, TypeError, KeyError):
    raise forms.ValidationError("(Hint: it's addition)")
```

## Not addressed here (deliberately)

The deeper cleanup — reading from `self.cleaned_data` instead of `self.data`, and fixing the `tries` default type mismatch — is described in #1136 but left for a follow-up to keep this PR small and easy to deploy.

## Test plan

- [ ] Submit registration form with `notarobot = '14.'` → form re-renders with hint message, no 500.
- [ ] Submit with correct numeric answer → registration proceeds as before.
- [ ] Submit with wrong numeric answer → form re-renders with hint (existing behavior).
- [ ] Confirm no admin error email is generated for the malformed case.

### DIFF
--- a/libraryauth/forms.py
+++ b/libraryauth/forms.py
@@ -96,7 +96,10 @@ class UserNamePass(UserData):
         return password2
 
     def clean_notarobot(self):
-        notarobot = int(self.data["notarobot"])
+        try:
+            notarobot = int(self.data["notarobot"])
+        except (ValueError, TypeError, KeyError):
+            raise forms.ValidationError("(Hint: it's addition)")
         encoded_answer = self.encode_answers.get(notarobot, 'miss')
         tries = self.data.get("tries", -1)
         if str(encoded_answer) != tries:


rdhyee:
Verified on test.unglue.it (production branch + this 4-line patch applied in-place, then reverted).

| Input | Before patch | After patch |
|---|---|---|
| `notarobot=14.` | HTTP 500 (bug repro) | HTTP 200, form re-renders with hint |
| `notarobot=hello` | would 500 | HTTP 200 |
| `notarobot=` (empty) | would 500 | HTTP 200 |
| `notarobot=999` (wrong int) | HTTP 200 with hint | HTTP 200 with hint (unchanged) |

No Python tracebacks in apache `error.log` during testing. Existing well-formed-submission behavior preserved.

Closes #1136.

PR #1143, opened 2026-05-07
Surface DOAB OAI 429 errors in load_doab cron output

rdhyee:
Cross-ref: [EbookFoundation/doab-check#12](https://github.com/EbookFoundation/doab-check/issues/12). **Why:** prod's outbound IP (52.4.17.140) is not whitelisted by DOAB while doab-check's (207.154.238.122) is — so prod hits HTTP 429s with 24h Retry-After every few days. Today's 04:30 UTC cron logged `loaded 0 records, ending at 2000-01-01`. Looked like a clean run; was actually a 429.

The Django logger has been faithfully recording these to `/var/log/regluit/unglue.it.log` since at least 2026-04-23, but operators monitor `/var/log/regluit/doab-harvest.log` (the cron's redirect target). The cron log is silent on 429s today.

## Diagnostic evidence

From `unglue.it.log`:
```
2026-04-23 15:24:10  ERROR  DOAB OAI rate-limited (HTTP 429). Retry-After: 86400 seconds. Harvest stopped after 482 records.
2026-05-06 04:36:51  ERROR  DOAB OAI rate-limited (HTTP 429). Retry-After: 86400 seconds. Harvest stopped after 1200 records.
2026-05-07 00:30:04  ERROR  DOAB OAI rate-limited (HTTP 429). Retry-After: 86400 seconds. Harvest stopped after 0 records.
```

The May 6 cron `loaded 1200 records (195 new)` line was a partial harvest — silently truncated by a 429 mid-stream. The May 7 cron `loaded 0 records` was a 429 on the very first batch.

## Change

- `core/loaders/doab.py`: `load_doab_oai` now returns `(num_doabs, new_doabs, lasttime, error)`. `error` is `None` on clean completion, a string like `'rate-limited (HTTP 429), retry-after=86400'` when the 429 path fires.
- `core/management/commands/load_doab.py`: prints `ERROR: DOAB OAI <error>` to stdout when present.

After this lands, the cron log will show:
```
starting at date:2026-05-04 00:00:00 until:None, max: 20000
ERROR: DOAB OAI rate-limited (HTTP 429), retry-after=86400
loaded 200 records (2 new), ending at 2026-05-06 20:08:12
```

## What this PR does NOT do

- **Does not fix the harvest.** The actual fix is whitelisting prod's IP at OAPEN — separately requested via email to Vaggelis Theodorakopoulos (the contact who whitelisted doab-check in March).
- **Does not surface bitstream-API 429s** (`doab_utils.get_streamdata`). Those fire per-record during ingest; surfacing them would need aggregation. Filed as follow-up if needed after the allowlist.
- **Does not retry.** Pure observability. If we want retry-with-backoff later, that's a separate decision.

## Test plan

- [ ] After deploy, confirm next 04:30 UTC cron run produces unchanged log line on success (no `ERROR:` line)
- [ ] If a 429 fires within a week, confirm `ERROR: DOAB OAI rate-limited...` appears in `doab-harvest.log`
- [ ] No unit test added — the diff is a pass-through return value; mocking pyoai's HTTPError path is more code than the fix

### DIFF
--- a/core/loaders/doab.py
+++ b/core/loaders/doab.py
@@ -1,6 +1,8 @@
 #!/usr/bin/env python
 # encoding: utf-8
+import collections
 import datetime
+import email.utils
 import logging
 import re
 import urllib.error
@@ -495,6 +497,7 @@ def load_doab_oai(from_date, until_date, limit=100):
     num_doabs = 0
     new_doabs = 0
     lasttime = datetime.datetime(2000, 1, 1)
+    error = None
     try:
         for record in doab_client.listRecords(metadataPrefix='oai_dc', from_=from_,
                                               until=until_date):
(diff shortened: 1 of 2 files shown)


rdhyee:
Review note: this improves stdout visibility, but the 429 path still exits successfully. That means cron/system monitors that key off exit status will continue to treat a failed or partial harvest as success, even though the command did not complete the requested harvest. If the operational contract is “harvest completed”, consider raising `CommandError` or otherwise returning non-zero after printing the ERROR line.

rdhyee:
Force-pushed a polish pass after a Codex review of the merge plan. The behavioral shape is unchanged — same 4-PR design Eric verbally endorsed. Improvements:

**Structured error data** — `load_doab_oai` now returns a `RateLimitedError` namedtuple `(kind, retry_after_seconds, retry_after_raw)` instead of a human-readable string. The mgmt command reads `.retry_after_seconds` directly; no more `error.split('retry-after=')`.

**RFC 9110 Retry-After parsing** — handles both forms per spec: delta-seconds (`86400`) AND HTTP-date (`Wed, 21 Oct 2099 07:28:00 GMT`). Done in the loader so all callers get parsed seconds.

**Fallback deadline** — if `Retry-After` is missing or unparseable, fall back to 1h (was: write no sentinel → retry immediately, which would defeat the whole patch).

**Timezone-aware UTC** — replaced deprecated `datetime.utcnow()` with `datetime.datetime.now(datetime.timezone.utc)`. Persisted ISO strings now carry the offset; legacy naive timestamps are accepted on read for backward compat.

**Loud sentinel-write failures** — silent `OSError` swallow replaced with stderr write + traceback. A permission/path failure would otherwise mask the very ban-extension loop this patch exists to prevent.

**Sentinel path** — moved from `/tmp/doab-oai-retry-after` (cleared on reboot, sometimes by tmpfiles age) to a durable location. Configurable via `DOAB_OAI_SENTINEL_PATH` env var for dev/test.

Tested locally: 9/9 cases for the Retry-After parser (delta-seconds, HTTP-date past/future, negative, whitespace, None, empty, malformed), and 7/7 cases for the read/write/clear sentinel roundtrip (including legacy naive ISO read, unwritable path returning False with stderr noise).

PR #1144, opened 2026-05-07
Respect DOAB OAI Retry-After across cron runs

rdhyee:
**Stacked on top of #1143** (review #1143 first; this PR's diff against `master` will look noisier until #1143 merges).

## Why

DOAB returns 429 with `Retry-After: 86400` (24 hours). Today we ignore that value: the cron runs again 24h later, hits 429 again, and DOAB sees a non-compliant client. This PR makes the harvest honor the server's instruction.

## Question for Eric / OAPEN (worth raising in the same email thread)

Do they consider unglue.it production conceptually the same trusted client as doab-check (which we operate as a service for OAPEN)? If yes, both IPs should be whitelisted as one logical client and this Retry-After honoring is just defense in depth. If no, we need to understand what they expect from prod's request profile — at which point reducing call rate would be the more substantive change.

## Change

When `load_doab_oai` returns a 429 error with a parseable Retry-After value, the management command:
1. Writes the deadline (UTC) to `/tmp/doab-oai-retry-after`
2. On any subsequent invocation, reads the sentinel and SKIPs if the current time is before the deadline
3. Clears the sentinel once the deadline passes

A new `--ignore-retry-after` flag lets operators override (e.g., after confirming with OAPEN that the limit was lifted).

## Cron log behavior after this lands

**During a ban window:**
```
SKIP: DOAB OAI rate-limited until 2026-05-08T00:30:04 UTC (honoring Retry-After). Use --ignore-retry-after to override.
```

**On the run that triggers the ban:**
```
starting at date:2026-05-04 00:00:00 until:None, max: 20000
Wrote Retry-After sentinel: blocked until 2026-05-08T04:30:00 UTC
ERROR: DOAB OAI rate-limited (HTTP 429), retry-after=86400
loaded 0 records (0 new), ending at 2000-01-01 00:00:00
```

## Why /tmp

Both regluit prod (cron under ubuntu) and dev environments need a writable path. `/tmp` is universally writable, persists across cron ticks, and survives short reboots — and DOAB's view of us survives reboots regardless, so a one-off sentinel loss after a long downtime would only mean one extra 429 before we re-honor.

Settings-driven would be cleaner but adds dev/prod settings work for a path that nobody will care to override.

## Test plan

- [ ] `manage.py load_doab 2026-05-04 --max=5` with no sentinel: runs normally
- [ ] After a 429: confirm `/tmp/doab-oai-retry-after` exists with future timestamp
- [ ] Re-run within the window: confirm SKIP message, no DOAB request made
- [ ] `--ignore-retry-after`: confirm bypasses SKIP
- [ ] After deadline passes: sentinel cleared on next run

### DIFF
--- a/core/management/commands/load_doab.py
+++ b/core/management/commands/load_doab.py
@@ -1,13 +1,76 @@
 import datetime
+import os
+import sys
+import traceback
+
 from django.core.management.base import BaseCommand
 
 from regluit.core.loaders import doab
 
+# Persisted across cron runs so we honor DOAB's Retry-After instead of
+# re-attempting inside the ban window (which extends the ban).
+#
+# Lives under /var/log/regluit because that directory already exists on prod,
+# is writable by the deploy user, and is excluded from the daily log-cleanup
+# cron (which targets `*.log` and `*.log.*` only). Override with the
+# DOAB_OAI_SENTINEL_PATH environment variable for non-prod environments.
+DEFAULT_SENTINEL = '/var/log/regluit/doab-oai-retry-after.state'
+SENTINEL_PATH = os.environ.get('DOAB_OAI_SENTINEL_PATH', DEFAULT_SENTINEL)
+
+
 def timefromiso(datestring):
     try:
         return datetime.datetime.strptime(datestring, "%Y-%m-%d")
-    except:
+    except ValueError:
         return datetime.datetime.strptime(datestring, "%Y-%m-%dT%H:%M:%S")
+
+
+def _now_utc():
+    return datetime.datetime.now(datetime.timezone.utc)
+
+
+def read_block_deadline(path=SENTINEL_PATH):
+    """Return the persisted Retry-After deadline (tz-aware UTC), or None."""
+    if not os.path.exists(path):
+        return None
+    try:
+        with open(path) as f:
+            raw = f.read().strip()
+        if not raw:
+            return None
+        parsed = datetime.datetime.fromisoformat(raw)
+    except (ValueError, OSError):
+        return None
+    if parsed.tzinfo is None:
+        parsed = parsed.replace(tzinfo=datetime.timezone.utc)
+    return parsed
+
+
+def write_block_deadline(deadline, path=SENTINEL_PATH, stderr=None):
+    """Persist the deadline; on failure, write a loud traceback to stderr.
+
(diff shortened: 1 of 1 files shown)


rdhyee:
Review note: `--ignore-retry-after` bypasses the sentinel for the current invocation, but a successful override run leaves the future `/tmp/doab-oai-retry-after` deadline in place. So if an operator uses the override after OAPEN confirms the block is lifted, the next normal cron run will still skip until the old deadline. Consider clearing the sentinel when `ignore` is set before running, or clearing it after a clean override run.

rdhyee:
Force-pushed a polish pass after a Codex review of the merge plan. The behavioral shape is unchanged — same 4-PR design Eric verbally endorsed. Improvements:

**Structured error data** — `load_doab_oai` now returns a `RateLimitedError` namedtuple `(kind, retry_after_seconds, retry_after_raw)` instead of a human-readable string. The mgmt command reads `.retry_after_seconds` directly; no more `error.split('retry-after=')`.

**RFC 9110 Retry-After parsing** — handles both forms per spec: delta-seconds (`86400`) AND HTTP-date (`Wed, 21 Oct 2099 07:28:00 GMT`). Done in the loader so all callers get parsed seconds.

**Fallback deadline** — if `Retry-After` is missing or unparseable, fall back to 1h (was: write no sentinel → retry immediately, which would defeat the whole patch).

**Timezone-aware UTC** — replaced deprecated `datetime.utcnow()` with `datetime.datetime.now(datetime.timezone.utc)`. Persisted ISO strings now carry the offset; legacy naive timestamps are accepted on read for backward compat.

**Loud sentinel-write failures** — silent `OSError` swallow replaced with stderr write + traceback. A permission/path failure would otherwise mask the very ban-extension loop this patch exists to prevent.

**Sentinel path** — moved from `/tmp/doab-oai-retry-after` (cleared on reboot, sometimes by tmpfiles age) to a durable location. Configurable via `DOAB_OAI_SENTINEL_PATH` env var for dev/test.

Tested locally: 9/9 cases for the Retry-After parser (delta-seconds, HTTP-date past/future, negative, whitespace, None, empty, malformed), and 7/7 cases for the read/write/clear sentinel roundtrip (including legacy naive ISO read, unwritable path returning False with stderr noise).

PR #1145, opened 2026-05-13
Fix auth_logout URL name in templates (Django 4.2)

rdhyee:
Surfaces against [#1123](https://github.com/Gluejar/regluit/pull/1123) — found during Google OAuth smoke-test on dj42.unglue.it today.

## Why

`{% url 'auth_logout' %}` in three templates was provided by the legacy `django-registration` package. After the upgrade to `django_registration 3.4` on this branch, that URL name no longer exists. `django.contrib.auth.urls` (already included in `libraryauth/urls.py:86`) provides the same view under the name `logout`.

Result: any authenticated request that renders `base.html` (the navbar's "Sign Out" link) crashes with `NoReverseMatch`. The page never gets to logged-in users because the navbar itself can't render.

## Reproduction on dj42

1. Sign in via Google OAuth (now functional after [regluit-provisioning#34](https://github.com/EbookFoundation/regluit-provisioning/pull/34))
2. Land back on `/`
3. → `Whoops, an error occurred`

Server log:
```
django.urls.exceptions.NoReverseMatch: Reverse for 'auth_logout' not found.
'auth_logout' is not a valid view function or pattern name.
  File "...frontend/templates/base.html", line 105 (rendered as base of /)
```

## Change

Rename `auth_logout` → `logout` in three templates:
- `frontend/templates/base.html:105` (navbar)
- `frontend/templates/pledge_user_error.html:15`
- `frontend/templates/gift_user_error.html:15`

## Known follow-up

Django 4.2's `LogoutView` accepts GET with a deprecation warning; Django 5.0 removes GET. The "Sign Out" link should eventually become a POST form. Out of scope here — this PR is the minimum patch to make any authenticated page render.

### DIFF
--- a/frontend/templates/base.html
+++ b/frontend/templates/base.html
@@ -102,7 +102,7 @@
                             {% if user.rights_holder.all %}
                             <li><a class="notbutton" href="{% url 'rightsholders' %}">Rights Holder Tools</a></li>
                             {% endif %}
-                            <li><a class="notbutton" href="{% url 'auth_logout' %}"><span>Sign Out</span></a></li>
+                            <li><a class="notbutton" href="{% url 'logout' %}"><span>Sign Out</span></a></li>
                         </ul>
                 </div>
             {% else %}
--- a/frontend/templates/gift_user_error.html
+++ b/frontend/templates/gift_user_error.html
@@ -12,7 +12,7 @@
 
 	<div><h2>Wrong user for Unglue.it credit</h2>
 		<div>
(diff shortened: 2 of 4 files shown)


rdhyee (commented):
Found one blocking issue: the new navbar link reverses to Django's `logout` view without passing `next`. In Django 4.2, `LogoutView` falls back to rendering `registration/logged_out.html` when there is no redirect target and no `LOGOUT_REDIRECT_URL`; this repo appears to have `frontend/templates/registration/logout.html`, not `registration/logged_out.html`. So authenticated users may trade the original `NoReverseMatch` for a `TemplateDoesNotExist` when clicking Sign Out. The two error-page links include `next` and avoid this path, but `frontend/templates/base.html:105` needs a redirect target, `LOGOUT_REDIRECT_URL`, or logout should be wired to the existing template.

rdhyee:
Codex review caught a follow-on gap. Pushed 452f75b to address:

> P1 navbar logout link has no \`next\` parameter. With Django 4.2's LogoutView, \`{% url 'logout' %}\` logs out on GET, but with no next and no \`LOGOUT_REDIRECT_URL\`, it falls through to rendering the default template \`registration/logged_out.html\`. This repo only has \`frontend/templates/registration/logout.html\`, so clicking "Sign Out" is likely to become a TemplateDoesNotExist 500.

Verified:
- \`LOGIN_REDIRECT_URL = "/"\` already set at \`settings/common.py:353\`; \`LOGOUT_REDIRECT_URL\` was not set
- \`frontend/templates/registration/logged_out.html\` does not exist; \`logout.html\` does

Fix: added \`LOGOUT_REDIRECT_URL = "/"\` symmetric with the existing LOGIN line. Logged-out users now redirect to the homepage (which renders cleanly for anonymous users post-template-fix). The vestigial \`logout.html\` template becomes dead weight — separate cleanup.

Codex's other notes confirmed.

PR #1152, opened 2026-05-14
Add backfill_doab management command (refs #1151)

### LINKED ISSUE #1151: DOAB harvest gap: ~20.6k active records missing (and ~7.3k stale records served) — coverage audit findings
## Why

Direct ID-level set-diff between unglue.it's `Identifier(type='doab')` rows and DOAB's authoritative OAI `ListIdentifiers` feed (run 2026-05-14, post-deploy of #1146/#1148/#1150) shows the rate-limited harvest period left a real coverage gap that the cron's 3-day rolling window cannot self-heal. The "we're caught up" signal from the cron log (clean runs since the deploy) is true *going forward* but doesn't reflect what was missed *during* the rate-limit window.

Related context: #1143, #1144, #1146, #1147, #1148, #1149, #1150 (the OAI 429 visibility / Retry-After / fork-pin / circuit-breaker series).

## Measurement

Pulled all 125,585 `<header>` entries from `https://directory.doabooks.org/oai/request?verb=ListIdentifiers&metadataPrefix=oai_dc` with status tracking (manual paginator, ~17 min, 0.3s pacing, Retry-After-respecting). Compared against `Identifier.objects.filter(type='doab').values_list('value', flat=True)` on prod.

```
unglue.it DOAB identifiers:        99,393
DOAB active (OAI, status != deleted):  110,792
DOAB deleted (OAI status = deleted):    14,793

=== ACTUAL GAP ===
Active in DOAB, missing from us:   20,608   (18.6% of active DOAB)
In us but DOAB marks deleted:       7,267   (stale records still served to users)
In us but absent from DOAB OAI:     1,942   (orphans — predate OAI feed?)
```

Note: the doab-check website's "99,451 books and book chapters" count and our `99,393` looked deceptively close in a totals-only comparison. They were close because the OAI feed includes ~15k deleted records that mask the real active-record gap. ID-level set-diff is the only reliable check.

## Distribution — where the loss happened

DOAB IDs are roughly chronological. Missing-records-as-percent-of-active by ID range:

| ID range | Missing | % gap |
|---|---|---|
| 20k–60k | 35–443 | **0.4–5.8%** (background noise) |
| 60k–70k | 371 | 4.4% |
| 70k–80k | 1,104 | 13.9% |
| **80k–100k** | **4,935** | **22–32%** |
| 100k–120k | 495 | 6–10% |
| 120k–140k | 3,061 | **25–27%** |
| 140k–150k | 804 | 12% |
| **150k–180k** | **9,095** | **34–48%** (most recent records) |

Pattern is unambiguous: gap is concentrated in newer record ranges. The recovery spikes visible in `/var/log/regluit/doab-harvest.log` (2026-04-23: 464 new, 2026-05-06: 478 new) helped but didn't fully catch up. Records dated in windows where every covering 3-day cron hit a 429 were lost from the incremental harvest until DOAB modifies them again.

### 1. Backfill ~20,608 active records missing from us

Concentrated in recent IDs, but spans 20k–179k. **Per established preference for slow & gentle DOAB harvesting** (we serve DOAB by directing users to DOAB-hosted content; looking like a misbehaving client undermines that), this should not be a single 20k-record blast.

**Proposed approach** (open to alternatives):
- One-off command that reads a list of missing DOAB IDs, fetches each via `getRecord` (not `listRecords`), small batches, Retry-After-respecting
- Pace at e.g. 200 records/day via cron → ~100 days to complete; or a one-time gentle run at e.g. 1 req/3sec → ~17h spread across off-peak hours
- Optionally restrict to records with high-value subsets first (e.g., post-2024 records only)

### 2. Retire ~7,267 stale (DOAB-deleted) records

These point to ebooks DOAB has withdrawn — we're showing users links to removed content. Purely local cleanup, no OAI traffic. Likely warrants its own decision (do we soft-delete? mark as withdrawn? keep history?). Probably belongs in a separate issue once approach is decided.

### 3. Investigate 1,942 orphans

`Identifier(type='doab')` rows whose value never appears in DOAB OAI. May predate the OAI feed, may be data-entry artifacts. Lower priority — may not be actionable.

## Follow-up: doab-check coverage is unmeasured (worth checking, not assumed similar)

`EbookFoundation/doab-check` runs on different infrastructure (DigitalOcean, separate IP from unglue.it) and has its own incident history independent of regluit's. The two services may have different gap shapes:

- **2026-03-11** ([doab-check#2](https://github.com/EbookFoundation/doab-check/issues/2)): DOAB issued a 24-hour IP-ban (`Retry-After: 86400`) against the doab-check harvester. The fix in that issue only *logged* the 429 — it didn't retry. Records modified during that 24-hour window were lost from the incremental harvest.
- **2026-05-05** ([doab-check#12](https://github.com/EbookFoundation/doab-check/issues/12)): silent cursor-stall bug — *not* a 429. The OAI resumption-token loop terminated after one page; cron returned ~98 records/run for 3 consecutive days while the cursor stayed pinned at `2026-05-01 06:05:25`. Manual catch-up loaded 1,520 records (146 new) for just May 1–5 alone. Workaround merged 2026-05-13 (commit `d680efd`).

Two distinct loss mechanisms, both unmeasured. The doab-check website's "65,620 links being checked" number is a derived count (links extracted from harvested records, not all records) and doesn't directly reveal coverage. **Recommend filing a parallel issue in EbookFoundation/doab-check to run the same set-diff** — the manual paginator script in this audit is reusable as-is. Don't assume the gap shape matches regluit's; measure separately.

## Reproducer

Manual paginator with state file (resumes on transient errors), saved at `/tmp/doab_diff3.py` on prod. Outputs:
- `/tmp/doab_oai_active.txt` — 110,792 active DOAB IDs
- `/tmp/doab_oai_deleted.txt` — 14,793 deleted DOAB IDs
- Set-diff script at `/tmp/doab_finaldiff.py`

Files are still on prod (`unglue.it`) for verification.

## Checklist

- [ ] Confirm gap with Eric; agree on backfill pacing
- [ ] Implement backfill command (separate PR)
- [ ] Decide treatment of 7,267 stale-deleted records (separate issue)
- [ ] File parallel issue in EbookFoundation/doab-check **and measure independently** (don't extrapolate from regluit)
- [ ] After backfill: re-run set-diff to confirm closure

rdhyee:
Closes #1151 once drained. Recovers the ~20.6k active DOAB records the nightly `load_doab` cron structurally cannot self-heal (3-day rolling window can't reach records modified during a 429 ban).

## What's in this PR (3 commits, bisectable)

| Commit | What |
|---|---|
| `2115772a` | `load_doab`: make the shared Retry-After sentinel monotonic + race-free (never-delete). A ban can now only be extended, never shortened or wiped by a racing process. Used by both the nightly harvest and the new backfill. |
| `88480de9` | `backfill_doab`: auto-discovery (Option B) + Codex-hardened reliability — removed `transaction.atomic()` around `add_by_doab` (the loader interleaves DB writes with up to ~3 min of HTTP per record; the nightly cron has run without an atomic wrapper for years). Precheck no longer freezes partial commits; 2nd 429 extends the sentinel; stock-pyoai fallback parses RFC 9110 Retry-After. |
| `2ff6226b` | `backfill_doab`: distinct exit codes `0/3/4` so the orchestration runner (separate PR EbookFoundation/regluit-provisioning#TBD) can honor the in-command `error_rate_halt` / `gone_rate_halt` circuit breakers ACROSS ticks (a naive re-run loop would defeat them). Every Retry-After SKIP path now `sys.exit(3)` so the runner does not falsely mark `.done` on the first ban. |

## Approval status

Code converged through **4 Codex review rounds** + **3 orchestration code-review rounds** (final: LGTM in both tracks). High-confidence on the code itself; the design calls below want human eyes (see "For Eric to weigh" — see the linked operator notes).

### Codex code-review trail (this PR + its `load_doab` dependency)

| Round | Outcome |
|---|---|
| R1 | Found 5: precheck freezing partial loads, 2nd-429 doesn't extend sentinel, stock-pyoai uses `int()` (mishandles HTTP-date), shared-sentinel races, discovery namespace-drift silently empty |
| R2 | 4/5 fixed; #4 (sentinel race) partial — `clear_sentinel()` safe, callers ignored refused-clear → fixed: returns live deadline, 5 call sites honor it |
| R3 | Caller fixes correct; deliberate `Retry-After: 0` → 60s non-fix endorsed (over-conservative, aligns with slow-and-gentle); root TOCTOU on `clear_sentinel`'s `os.remove` → fixed per Codex's own prescription: never delete (an expired sentinel is harmless; next 429 overwrites; monotonic write guarantees correctness) |
| **R4** | **LGTM** — all caller fixes verified, contract satisfied, no caller depended on deletion, no growth/staleness, both repos consistent |

### Design calls that deserve maintainer eyes (not Codex's domain)

1. **Removing `transaction.atomic()` around `add_by_doab`** in the regluit loader. Justified by the loader's HTTP-in-DB-transaction problem (cover fetch up to ~120s) + the established prod cron pattern (`load_doab_oai` has never wrapped `add_by_doab` in a transaction). The precheck change (commit `88480de9`) depends on `add_by_doab`'s idempotency to heal any partial commits — please sanity-check that idempotency claim against the cover/metadata/ISBN follow-up code paths.

2. **`load_doab.py` sentinel change** (commit `2115772a`) touches the **live nightly harvester's** shared helpers. The change is *strictly safer* (bans can only be extended; no `os.remove` race), but it is production-cron-affecting and so wants a careful read.

## What this enables but does not deliver

- **Orchestration to actually drain the 20.6k**: see the companion PR in EbookFoundation/regluit-provisioning (`feature/doab-backfill-orchestration`) — adds the runner script under shared `flock` with the nightly harvest, marker-file circuit breaker, `deploy_type == 'prod'`-gated cron. The exit-code change in this PR is what makes that runner safe.
- **Tests** for `backfill_doab` itself: deliberately out of scope for this PR (the command is straightforward to exercise by hand; the design budget went to Codex hardening + the orchestration). Tracked for follow-up.

## Backlinks

- Issue: #1151 (coverage gap audit, ~20.6k missing active records)
- Companion: EbookFoundation/regluit-provisioning#TBD (orchestration)
- Companion: EbookFoundation/doab-check#19 (parallel doab-check backfill for the ~14.5k there — divergence-aware port)

🤖 Generated with [Claude Code](https://claude.com/claude-code) (Opus 4.7, 1M context) — multi-round Codex CLI adversarial review

### DIFF
--- a/core/management/commands/backfill_doab.py
+++ b/core/management/commands/backfill_doab.py
@@ -3,12 +3,15 @@ backfill_doab — gentle, resumable, observable DOAB ID-list backfill.
 
 One-off catch-up command for IDs the nightly `load_doab` cron missed (because
 its 3-day rolling window can't self-heal records modified during a stretch of
-HTTP 429s). Reads a static list of DOAB IDs, processes them with rate-limited,
-ID-keyed state, and a 4-way per-record outcome taxonomy.
+HTTP 429s). The worklist is auto-discovered (DOAB OAI ListIdentifiers diffed
+against the local DB) when --ids-file is omitted, or supplied explicitly for
+targeted re-runs. IDs are processed with rate-limited, ID-keyed state and a
+4-way per-record outcome taxonomy.
 
-Design context: Gluejar/regluit#1151. Reuses `add_by_doab` (idempotent) and the
-shared Retry-After sentinel from `load_doab.py` so a 429 hit by either command
-suppresses the other for the duration of the ban.
+Design context: Gluejar/regluit#1151. Reuses `add_by_doab` (idempotent), the
+existing pyoai client / OAI-prefix stripper, and the shared Retry-After
+sentinel from `load_doab.py` so a 429 hit by discovery, the per-record loop,
+or the nightly cron suppresses the others for the duration of the ban.
 """
 
 import argparse
@@ -25,7 +28,7 @@ import traceback
 import urllib.error
 
 from django.core.management.base import BaseCommand, CommandError
-from django.db import IntegrityError, transaction
+from django.db import IntegrityError
 
 from regluit.core.loaders import doab
 from regluit.core.management.commands.load_doab import (
@@ -43,6 +46,8 @@ except ImportError:
         retry_after_seconds = None
         retry_after_raw = None
 
+from oaipmh.error import NoRecordsMatchError
+
 logger = logging.getLogger(__name__)
 
 # Terminal statuses (won't re-attempt)
@@ -56,16 +61,6 @@ ID_RE = re.compile(r'^20\.500\.12854/(\d+)(?:\.\d+)?$')
 
 
 # --- ID parsing ---------------------------------------------------------------
-
-def parse_id_line(line):
-    """Validate and return a normalized DOAB ID, or None if malformed."""
-    s = line.strip()
-    if not s or s.startswith('#'):
-        return None
-    if not ID_RE.match(s):
-        return s if s else None  # caller decides whether to reject
-    return s
-
 
 def numeric_suffix(doab_id):
     """Parse the integer suffix after `20.500.12854/` for sorting."""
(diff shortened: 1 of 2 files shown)


rdhyee (commented):
Code review findings:

- High: the command continues processing after a short 429. `_handle_rate_limit()` returns `recovered`, the caller retries the record, and then the loop continues to the next pending ID. That violates the stated first-429 stop condition and can keep calling DOAB after the shared sentinel has been written. I would checkpoint the current ID and exit after the courtesy retry, or remove the courtesy retry and always exit on 429.

- Medium: the transaction still wraps remote work. The command fetches the OAI record before `transaction.atomic()`, but `doab.add_by_doab()` can still make cover/bitstream HTTP calls via `update_cover_doab()` / `store_doab_cover()` / `get_streamdata()`. That means the DB transaction can be held during HTTP despite the design note. Either move those remote side effects out of the atomic block, add a loader path that skips cover fetching during backfill, or narrow the transaction around only the actual DB writes.

- Medium: deleted OAI records are not categorized as `gone`. A deleted/no-content record gets to `add_by_doab()`, returns `None`, and the command marks it `error_review`. That inflates review errors and bypasses the intended gone-rate halt. The command should detect deleted/no-content OAI headers before calling `add_by_doab()` and mark those as `gone`.

I did not run tests; this was a code review against the loader contracts and the PR diff.

rdhyee:
**Addressed codex review findings** in commit e103d020:

| Codex finding | Severity | How addressed |
|---|---|---|
| 429-recovered path falls through, continues processing | HIGH | After single retry, set `stop_after_this_record=True`, break after categorizing the result. All `continue` paths within the loop check the flag. |
| Deleted OAI records marked `error_review` instead of `gone` | MEDIUM | New step in 7d: check `record[0].isDeleted() or not record[1]` before calling `add_by_doab()`; mark `gone` directly. |
| `transaction.atomic()` "scoped to local writes only" claim is false | MEDIUM | Replaced misleading comment with honest trade-off explanation. The wrap *does* hold a DB tx during downstream HTTP calls inside `add_by_doab` (cover image, bitstream). Refactoring `add_by_doab` to support fetch-vs-write phase separation is out of scope for this PR. |

The 4th codex finding (status flip) only applies to doab-check; addressed in EbookFoundation/doab-check#19.

rdhyee:
Pass-2 codex finding addressed in 3c148828: stock-pyoai HTTPError(429) fallback now writes sentinel + marks 'retry' + exits cleanly. No sleep, no retry attempt — keeps that behavior exclusively in the patched-fork path which already has the stop_after_this_record handling.

rdhyee (commented):
LGTM from this review pass. The previous 429 fallback issue is resolved, the updated command compiles, and I do not see remaining blocking findings.

Residual caveat: tests and a small capped smoke run are still important before taking this out of draft / running at production scale.

- codex

rdhyee:
2026-05-20 status update after the cross-host snapshot work:

- Latest head includes `cfe054e2`, which adds `.active` / `.deleted` snapshot emission plus `--use-active-file` / `--use-deleted-file` consumer mode. The original PR body is now stale where it says "3 commits" and where it still names the provisioning companion as TBD.
- Test-machine demo path now works end-to-end: one DOAB OAI crawl on `test.unglue.it` produced the universal `.active` / `.deleted` files plus the local `.ids` / `.stale` / `.orphans` artifacts; reusing `.active` on the same host produced a byte-identical worklist with zero DOAB calls.
- Cross-codebase reuse has also been proven outside this PR: the same `.active` snapshot was consumed by a throwaway doab-check diff script, producing a 14,495-line ids file in seconds, and doab-check's existing `backfill_doab` successfully loaded one test record from that worklist.
- This PR may still be simplified before merge. The producer half (emit `.active` / `.deleted`) is small and clearly useful. The consumer half (`--use-active-file` flags) is reusable, but if this remains a one-off backfill we may trim to producer-only and keep cross-host consumption as an operator script.

Operational caveat surfaced by the demo: DOAB ID `20.500.12854/176562` loaded successfully in doab-check but failed through regluit's `add_by_doab` path. That looks like a pre-existing regluit loader divergence and should be tracked separately rather than blocking the backfill-scope decision.

PR #1158, opened 2026-06-08
Fix #1156: widget endpoint 500s on unknown/non-numeric/invalid ids

### LINKED ISSUE #1156: DoesNotExist 500 on /api/widget/<id>/: unguarded Work.DoesNotExist from safe_get_work for non-numeric/unknown ids
## Summary

`GET /api/widget/featured4iL7gEXx/` returned a 500:

```
DoesNotExist at /api/widget/featured4iL7gEXx/
(Work matching query does not exist)
```

Exception origin: `api/views.py` `widget()` → `core/models/bibmodels.py:1082` `safe_get_work()` raising `Work.DoesNotExist()`.

## Root cause

The `widget` view (`api/views.py:59`):

```python
def widget(request, isbn):
    if isbn == 'featured':                       # 64: exact match only
        work = featured_work()
    else:
        if len(isbn) == 10:
            isbn = regluit.core.isbn.convert_10_to_13(isbn)   # 67: can return None
        if len(isbn) == 13:                                   # 69: len(None) -> TypeError
            try:
                identifier = models.Identifier.objects.get(type='isbn', value=isbn)
                work = identifier.work
            except models.Identifier.DoesNotExist:
                return render(request, 'widget.html', {'work': None})
        else:
            work = models.safe_get_work(isbn)    # 78: can raise Work.DoesNotExist, uncaught
    return render(request, 'widget.html', {'work': work})
```

For `isbn = 'featured4iL7gEXx'`: not `'featured'`, length neither 10 nor 13, so it falls to `safe_get_work('featured4iL7gEXx')`. That id is non-numeric → `safe_get_work` hits `except ValueError: raise Work.DoesNotExist()` (`bibmodels.py:1080-1082`). The view does not catch it → 500.

The view already handles the analogous "not found" case for the ISBN branch (`Identifier.DoesNotExist` → render with `work=None`); the `safe_get_work` branch lacks the same guard.

## Two related defects in the same view (found via code review)

1. **Template variable missing.** `widget.html:41` renders `No work corresponding to ISBN {{isbn}} available`, but the `work=None` render paths do **not** pass `isbn` into the context. The empty-widget message therefore drops the identifier. Pass `isbn` along when rendering the not-found widget.
2. **Invalid 10-char ISBN → TypeError.** `convert_10_to_13` returns `None` for an invalid ISBN (`core/isbn.py:44-46`). The next line does `if len(isbn) == 13`, i.e. `len(None)` → `TypeError` → another 500. A 10-character bogus token (e.g. `/api/widget/ABCDEFGHIJ/`) triggers this.

## Proposed fix

- Wrap `safe_get_work(isbn)` in `try/except models.Work.DoesNotExist:` → render `widget.html` with `work=None` (and `isbn`).
- After `convert_10_to_13`, if it returns `None`, render the empty widget instead of calling `len(None)`.
- Pass `isbn` into all `work=None` render paths so the template message renders.

## Tests

- `GET /api/widget/featured4iL7gEXx/` → 200, `work=None` (was 500).
- `GET /api/widget/999999999/` (numeric, no such work) → 200, `work=None`.
- `GET /api/widget/ABCDEFGHIJ/` (invalid 10-char) → 200 (was `TypeError` 500).

## Affected code

- `api/views.py` — `widget()` (≈ 59–82)
- `api/templates/widget.html:41` (context dependency)

## Notes

- The id `featured4iL7gEXx` looks like a garbled/probing request (`'featured'` + junk) — likely a broken embed or a scanner. A graceful empty widget is the correct behavior.
- Present on `production` (verified on the deployed branch).

---
*Diagnosis assisted by Claude Code; reviewed and refined with OpenAI Codex (gpt-5.4).*

rdhyee:
Fixes #1156 — `/api/widget/<id>/` returns 500 for unknown, non-numeric, or invalid identifiers instead of the existing empty-widget response.

## Why

The `widget` view treats any non-`featured`, non-10/13-character token as a work id and calls `safe_get_work(isbn)`. For a token like `featured4iL7gEXx`, `safe_get_work` converts Django's integer-field `ValueError` into `Work.DoesNotExist`, which the view never catches → 500. Two adjacent defects compound it:
- `convert_10_to_13()` returns `None` for an invalid ISBN-10, so the following `len(isbn)` raised `TypeError` (another 500) for any bogus 10-char token.
- `widget.html` renders `...ISBN {{isbn}}...`, but the `work=None` render paths didn't pass `isbn`, so the empty-widget message dropped the identifier.

## What changed (`api/views.py`)

- Wrap `safe_get_work(isbn)` in `try/except models.Work.DoesNotExist` → render the empty widget, matching the existing `Identifier.DoesNotExist` handling.
- Guard the ISBN branch with `if isbn and len(isbn) == 13` so a `None` from `convert_10_to_13` falls through to the empty-widget response instead of `len(None)`.
- Pass `isbn` into every `work=None` render path so the template message renders.

## Tests (`api/tests.py`, `ApiTests`)

- `/api/widget/featured4iL7gEXx/` → 200 (was 500)
- `/api/widget/999999999/` (numeric, no such work) → 200
- `/api/widget/ABCDEFGHIJ/` (invalid ISBN-10) → 200 (was `TypeError` 500)

## Validation

`py_compile` clean; `convert_10_to_13` confirmed to return `None` (not raise) on invalid input. Diagnosis reviewed and refined with OpenAI Codex (gpt-5.4).

### DIFF
--- a/api/tests.py
+++ b/api/tests.py
@@ -50,6 +50,19 @@ class ApiTests(TestCase):
         r = self.client.get('/api/widget/%s/'%self.work_id)
         self.assertEqual(r.status_code, 200)
 
+    def test_widget_unknown_ids_return_empty_widget(self):
+        # Regression for #1156: bad identifiers must render an empty widget (200),
+        # not raise an uncaught exception (500).
+        # non-numeric token (length != 10/13) -> safe_get_work raises Work.DoesNotExist
+        r = self.client.get('/api/widget/featured4iL7gEXx/')
+        self.assertEqual(r.status_code, 200)
+        # numeric but no such work
+        r = self.client.get('/api/widget/999999999/')
+        self.assertEqual(r.status_code, 200)
+        # invalid ISBN-10 -> convert_10_to_13 returns None (was len(None) TypeError)
+        r = self.client.get('/api/widget/ABCDEFGHIJ/')
+        self.assertEqual(r.status_code, 200)
+
 class FeedTests(TestCase):
     fixtures = ['initial_data.json', 'neuromancer.json']
     def setUp(self):
--- a/api/views.py
+++ b/api/views.py
@@ -65,19 +65,29 @@ def widget(request, isbn):
         work = featured_work()
     else :
         if len(isbn)==10:
+            # convert_10_to_13 returns None for an invalid ISBN-10; guard the
+            # len() below and fall through to the empty-widget response.
             isbn = regluit.core.isbn.convert_10_to_13(isbn)
-        if len(isbn)==13:
+        if isbn and len(isbn)==13:
             try:
                 identifier = models.Identifier.objects.get(type = 'isbn', value = isbn )
                 work = identifier.work
             except models.Identifier.DoesNotExist:
                 return render(request, 'widget.html',
-                     { 'work':None,},
+                     { 'work':None, 'isbn':isbn, },
                     )
         else:
-            work= models.safe_get_work(isbn)
+            # isbn may be a work_id, a non-numeric token, or None (invalid
+            # ISBN-10). safe_get_work raises Work.DoesNotExist for unknown or
+            # non-numeric ids; render an empty widget instead of a 500.
+            try:
+                work = models.safe_get_work(isbn)
+            except models.Work.DoesNotExist:
+                return render(request, 'widget.html',
+                     { 'work':None, 'isbn':isbn, },
+                    )
     return render(request, 'widget.html',
-         {'work':work, },
+         {'work':work, 'isbn':isbn, },
      )
 
 def featured_cover(request):


rdhyee:
### Raymond asked

> is the widget supposed to accept non-ISBNs at all -- that's my first gut reaction

### Claude answered

The parameter is *named* `isbn` but the endpoint was never ISBN-only. The site itself generates three different argument shapes:

1. **ISBN-13** — `libraries.html`, `pledge_complete.html`, `download.html` embed `/api/widget/{{work.first_isbn_13}}/`
2. **Work id** — `work_action.html` embeds `/api/widget/{{work.id}}/` ← this is why the final `else` branch calls `safe_get_work` (work-id lookup) at all
3. **The literal `featured`** — `api_help.html` advertises `/api/widget/featured/` as the rotating featured-book widget

So `isbn` is a misnomer for "isbn-or-work_id-or-'featured'", and the route regex `\w+` admits all of them.

Two consequences for review:

**The route can't be tightened.** These are copy-paste embed codes published into third-party webpages for over a decade. Whatever URLs were ever handed out must keep working — narrowing the route to ISBN-only would break every site that embedded a work-id widget. So the PR's choice — accept the junk, render an empty widget — is the right call.

**A pleasing detail:** the actual crashing input from the error email — `featured4iL7gEXx` — reads like `featured` + line noise: most plausibly a *mangled embed* of the featured-widget code (broken copy-paste, or a bot mutating URLs). Length 16 → falls past both ISBN branches → `safe_get_work('featured4iL7gEXx')` → `Work.DoesNotExist` → the old 500. Under the fix, that broken embed renders an empty widget box instead of an iframe full of error — arguably the *correct* user-visible outcome for a corrupted embed.

(Parked for the bot-protection workstream, not this hotfix: renaming the parameter `isbn` → `identifier`, and rate-limiting / `noindex`-ing the widget route.)

### Raymond asked

> So given that design, the gist is that we didn't have clean logic to handle all 3 cases?

### Claude answered

Right — with one sharpening: the old logic *dispatched* the three cases fine; what it lacked was handling for **invalid members** of each case. It was happy-path-only code. Lay out the full input lattice and the holes become obvious:

| Input | Old code path | Valid input | Invalid input |
|---|---|---|---|
| `featured` | `featured_work()` | ✅ works | (n/a — literal match) |
| len 10 | `convert_10_to_13` → fall through | ✅ valid ISBN-10 converts, proceeds | ❌ **invalid ISBN-10 → `None` → `len(None)` → 500** |
| len 13 | `Identifier.objects.get` in try/except | ✅ found → widget | ✅ not found → caught, empty widget |
| everything else | `safe_get_work(isbn)` bare | ✅ real work id → widget | ❌ **unknown id / junk token → `Work.DoesNotExist` uncaught → 500** |

Note the asymmetry in the third row: whoever wrote the ISBN-13 lookup *did* think about the not-found case — there's a try/except right there returning the empty widget. The work-id branch, doing the morally identical lookup two lines below, got no such protection. And the ISBN-10 branch assumed conversion always succeeds. So it's not that the design was confused — it's that defensive handling was applied to exactly one of the three lookup paths and forgotten on the others.

The PR's fix is, in essence, *completing the lattice*: every cell now terminates in either a real widget or the empty widget, never an exception. One-sentence summary: **the widget endpoint had three input types but graceful-failure handling on only one of them; #1158 extends the existing not-found behavior to the other two.**

It also explains why this sat latent for ~14 years: humans copy-pasting embed codes produce valid inputs; it took crawlers feeding mutated URLs (`featured4iL7gEXx`) at scale to systematically explore the invalid cells.

-- Raymond asked; Claude answered

rdhyee:
## Test evidence: red→green on both stacks (2026-06-10)

Verifies #1156's fix empirically — first actual execution of this PR's test (previously validated only via py_compile + logic replay + Codex review). Same method as the [#1157 evidence run](https://github.com/Gluejar/regluit/pull/1157#issuecomment-4671965690); the full scratch-checkout recipe and gotchas are documented there and apply verbatim (substitute this PR's branch).

| Run | Box / stack | Code under test | Result |
|---|---|---|---|
| **RED** | test.unglue.it · Django 1.11.29 / Py 3.8 | `master` (`42a984c0`) old `api/views.py` + this PR's test | **FAILED (errors=1)** — expected |
| **GREEN** | test.unglue.it · Django 1.11.29 / Py 3.8 | this branch (`1f82b618`) | **OK** (full `ApiTests`) |
| **GREEN** | dj42.unglue.it · Django 4.2.21 / Py 3.12 | `1f82b618` cherry-picked onto `fix/django-upgrade-libraryauth` — applied clean, no conflicts | **OK** |

### Why the RED run matters

Unlike #1157 (a race that single-threaded tests can't trigger), these bugs are deterministic — so the red run **reproduces the production failure exactly**. The traceback is the one from the forwarded error email:

```
File "/tmp/pr1157/regluit/api/views.py", line 78, in widget
    work= models.safe_get_work(isbn)
File "/tmp/pr1157/regluit/core/models/bibmodels.py", line 1082, in safe_get_work
    raise Work.DoesNotExist()
regluit.core.models.bibmodels.DoesNotExist
```

i.e. `GET /api/widget/featured4iL7gEXx/` → uncaught `Work.DoesNotExist` → 500, on the old code; 200 + empty widget on the fixed code. (The run errors at the first of the three junk URLs in the test, which documents the headline defect; the `len(None)` invalid-ISBN-10 path is exercised by the same test once the code lets execution proceed that far — covered in the GREEN runs.)

### Notes

- On dj42 the cherry-pick was applied **on top of the #1157 cherry-picks** — so the green run there also shows the two hotfixes coexist cleanly on the Django 4.2 branch.
- Reproduce by hand: follow steps 1–4 of the [#1157 comment](https://github.com/Gluejar/regluit/pull/1157#issuecomment-4671965690), replacing the branch with `fix/1156-widget-doesnotexist` and the test label with `regluit.api.tests.ApiTests`. The scratch checkouts at `/tmp/pr1157/` on both boxes are still in place.

---
*Runs executed against throwaway clones + test databases by Claude Code, 2026-06-10; live deployments untouched; results reported verbatim.*

eshellman:
the invalid strings should result in a 404 not a 500 or 200. but yes, the widget should accept non-isbn. values

Issue #1172, opened 2026-06-18
unglue.it (new Django 4.2 box) CPU-saturated by AI crawlers — load 27 on 4 cores (ClaudeBot = 26% of traffic)


rdhyee:
## Why / summary

Since the 2026-06-17 cutover, the new green box serving unglue.it is **running near 100% CPU with no headroom** and load has been *climbing* all night. The site is still serving correctly (200s dominate, only 3 HTTP 500s all day, no instance failures), but it has no margin — uncomfortably close to the "wedges under load" mode we left the old stack to escape. Filing so we fix it deliberately and loop Eric. **Note: bot-protection work already exists unmerged (`bot-protection-middleware`, `bot-protection-downloads`) — see Recommendations.**

## Evidence (live diagnostics, green `i-0b0d4da7db10adee2`, 2026-06-18 ~14:43 UTC)

- **Load average `27.5 / 18.7 / 12.2`** (1/5/15-min) on a **4-core** box (~7× oversubscribed), **trending up**.
- **CPU**: `84% us, 0% idle`. Hot processes are all **apache2 / mod_wsgi (Django)** workers, each pinned 58–100%. **Celery is quiet** (4–9%).
- **Memory healthy**: 8.3 GB free of 15.7 GB, **swap=0, no swapping** → genuine compute, *not* a memory leak.
- **`DEBUG=False` confirmed** for prod (`regluit.settings.prod`); the `DEBUG=True` in the tree is `dev.py` only. Not a factor, not a security issue.
- **mod_wsgi `processes=2 threads=15`** → GIL-bound; the Python tier is the bottleneck.
- unglue.it terminates TLS on-box (certbot + EIP) — **not currently proxied through Cloudflare/a CDN.**

### Traffic (today's access log, ~14h, 476,413 requests)

| Client | Requests | % | Identity |
|---|---|---|---|
| `216.73.216.186` | **124,023** | **26%** | **`ClaudeBot/1.0` (claudebot@anthropic.com)** |
| `74.7.243.222` | 23,701 | 5% | `GPTBot/1.4` (OpenAI) |
| `57.141.0.x` (many IPs) | ~2,000 each | — | Meta-range crawl farm |
| `136.243.228.194` | 7,778 | — | Hetzner (datacenter) |
| `66.249.70.71` | 6,349 | — | Googlebot (legit, keep) |

- **AI training crawlers dominate** — ClaudeBot alone is **26% of all traffic**, hammering `/work/` pages.
[...]

PR #1116, opened 2026-03-26
Remove faceted search UI — Phase 1 (#1110)

rdhyee:
**Refs #1110, #1095** — Subjects (457K values) × other facets = hundreds of millions of crawlable compound URLs that bots exploit, driving server load.

## Summary

Blocks keyword/subject facets from combining with other facet types. Compound keyword URLs return 404 immediately instead of running expensive DB queries. All single-facet browsing and non-keyword compounds preserved.

### Changes (6 files)

**`core/facets.py`**:
- `get_other_groups()`: keyword active → no refinement offered; non-keyword active → keywords excluded
- `get_facet_object()`: raises `InvalidFacetCombination` for keyword + other compounds

**`frontend/views/__init__.py`**: catches `InvalidFacetCombination` → 404

**`api/views.py`**: catches `InvalidFacetCombination` in OPDS, OPDS-JSON, and ONIX views → 404

**`frontend/templates/explore.html`**: removed per-subject links from sidebar

**`frontend/tests.py`**: 7 tests for web UI behavior

**`api/tests.py`**: 4 tests for OPDS/OPDS-JSON/ONIX 404 behavior

### Behavior matrix

| Page | Keywords offered? | Other facets offered? | Status |
|------|------------------|----------------------|--------|
| `/free/` | Yes | Yes | 200 |
| `/free/kw.Fiction/` | N/A | **No** — blocked | 200 |
| `/free/kw.Fiction/epub/` | — | — | **404** |
| `/free/epub/kw.Fiction/` | — | — | **404** |
| `/free/epub/` | **No** — blocked | Yes (minus Format) | 200 |
| `/free/epub/doab/` | **No** — blocked | Yes (remaining) | 200 |
| `/api/opds/kw.Fiction/epub/` | — | — | **404** |
| `/api/onix/kw.Fiction/epub/` | — | — | **404** |

### 13 tests passing on test.unglue.it

```
Frontend: 7/7 ok
OPDS/OPDS-JSON/ONIX: 6/6 ok (including 4 new + 2 existing)
```

### Review history
- Multiple rounds of Codex, Gemini, OpenCode review
- Each round caught real issues (unconditional keyword exclusion, missing ONIX handler, import error, test gaps)
- All fixed and verified

PR #1173, opened 2026-06-18
robots.txt: throttle ClaudeBot, disallow training crawlers, keep search agents (refs #1172)

rdhyee:
## In plain English

**What this does.** Edits one file — the `robots.txt` we serve — to ask AI crawlers to slow down or stay away, while leaving search engines alone.

**Benefit.** On 2026-09-02, crawler traffic pushed the site into returning errors for about seventeen minutes. This asks the heaviest crawler to pause between requests and skip our most expensive pages, and tells the training-only crawlers not to come at all. It should make those episodes rarer, not impossible: `robots.txt` is a request, not a fence. Crawlers that ignore it are unaffected, and those that honor it only pick up the change on their next fetch — which can take a day.

**Risk.** The realistic failure is excluding something we wanted found. Mis-filing a crawler costs search visibility while saving no load; review caught exactly that twice and both entries were removed. Search engines are deliberately untouched. Separately, a **pre-existing** quirk: the permissive rules apply only to the exact host `unglue.it`, so `www.unglue.it` is told to stay out. **This PR leaves that unchanged** — it predates this work and needs its own fix.

**How we'd know.** After deploy, fetch `/robots.txt` and check it reads right. Over the next few days, watch whether those crawlers' request volume drops and the error bursts stop recurring.

**Rollback.** Revert one file.

---

Refs #1172. Adds AI-crawler handling to the served `robots.txt` — the *polite* layer. Opened as a discussion draft in June; now a concrete proposal, with the decisions it raised resolved, a regression test added, and five rounds of adversarial review applied.

## What changed since the draft

**1. Fixed a group-precedence bug in the draft's own rules.** The `User-agent: *` group applies only to crawlers that match no named group. The draft's `ClaudeBot` group carried only `Crawl-delay`, so shipping it would have **removed** ClaudeBot's existing `Disallow` rules for `/accounts/`, `/feedback/`, `/socialauth/`, `/search/`, `/googlebooks/` — widening its access into the faceted-search URL space that #1116 was cut to eliminate, rather than narrowing it. Every named group that isn't a blanket `Disallow: /` now restates the baseline.

**2. Verified the Crawl-delay premise instead of asserting it.** Anthropic's published crawler documentation states they "support the non-standard Crawl-delay extension to robots.txt" and gives a `User-agent: ClaudeBot / Crawl-delay: N` example. So throttling ClaudeBot is meaningful rather than decorative. GPTBot and the other training crawlers document no such support, so there is nothing to throttle and they stay disallowed.

**3. ClaudeBot also excludes the expensive endpoints.** Beyond the baseline, its group disallows `/free/`, `/bypub/`, `/pid/`, `/unglued/`, `/campaigns/` and `/api/` — the bot-hit listing and feed paths identified in #1189 — while leaving `/work/` pages crawlable. This is the actual load-shedding; the throttle sheds it without costing discoverability of the book content.

**4. Removed entries that cost discoverability without shedding load.** Two rounds of review found the same shape twice, both inherited from the draft's original list:

| Entry | Why it's gone |
|---|---|
| `PerplexityBot` | Perplexity documents it as their **search-indexing** crawler, explicitly "not used to crawl content for AI foundation models." Blocking it would drop the collection out of Perplexity search results while shedding no training crawl at all. |
| `Google-Extended`, `Applebot-Extended` | **Not crawlers.** Googlebot and Applebot do the fetching; these are content-*usage* opt-out tokens. They shed zero crawl load, while `Google-Extended` also costs Gemini grounding. Whether to opt out of AI training by the search engines is a content-policy question worth deciding on its own terms — not something to fold into a load change. |

`Amazonbot` was checked the same way and **stays** disallowed: Amazon documents it as crawling that "may be used to train Amazon AI models." Their separate `Amzn-SearchBot` does not, and already falls through untouched.

`cohere-ai` was also dropped — Cohere's documentation states they operate no crawler for training generative models "at this time" and names no active agent, so the entry sheds no load and implied a protection that doesn't exist.

**One entry worth a deliberate decision:** `CCBot` stays disallowed, but note that Common Crawl **does** honor `Crawl-delay`. So unlike the others in that block, blocking it is a content-policy choice — opting out of the Common Crawl dataset — rather than a technical necessity. It could be throttled like ClaudeBot instead. Flagged rather than decided.

**5. Cleaned up the served file.** The inline `DECISION FOR REVIEW` comments are gone now that the decisions are made, and the traffic statistics have been dropped — this template is public output, so it carries policy, not operational detail.

## Decisions resolved

| Question from the draft | Resolution |
|---|---|
| ClaudeBot: throttle vs full Disallow? | **Throttle** (`Crawl-delay: 30`) plus the expensive-path exclusions — keeps book pages discoverable |
| Keep AI *search* agents allowed? | **Yes.** `PerplexityBot`, `OAI-SearchBot`, `Claude-SearchBot`, `Amzn-SearchBot` and the user-triggered fetchers get no group and fall through to `*` |
| Googlebot | Untouched |

## Tests

New `RobotsTxtTests` in `frontend/tests.py`, parsing the rendered file rather than pattern-matching it:

- production host serves the baseline plus the AI rules, `Content-Type: text/plain`;
- ClaudeBot carries the `Crawl-delay` **and** every baseline and expensive-path exclusion; every listed training crawler is genuinely blocked; and the set of groups is asserted exhaustively, so a stanza cannot be dropped or slipped in unnoticed;
- search and user-triggered agents have **no** group of their own, so a future edit can't quietly start blocking one;
- non-production hosts (staging, localhost) still serve a blanket `Disallow: /`, so non-canonical copies stay unindexed;
- **a regression guard** asserting no named group omits the baseline rules — i.e. that adding a group can never widen access;
- a fixture for the parser itself, so an extension record between two `User-agent` lines can't silently split a group and drop an agent out of the guard's coverage.

These run as a `TestCase`, not a `SimpleTestCase`: issuing the request loads the root URLconf, which imports a module that runs a query at import time, so a `SimpleTestCase` (which builds no test database) would have let that query reach whatever database the settings point at. The real URLconf is kept rather than a minimal stand-in, so the test still proves `/robots.txt` is actually routed.

Each guard was checked by breaking the file on purpose and confirming the suite fails: reverting to the pre-fix template trips the precedence guard, and deleting `/api/` or the `Bytespider` stanza trips the new exhaustive assertions.

Full `frontend` suite locally: 41 tests, 3 failures — all three pre-existing (`RhPageTests`), confirmed by running the same suite with this change stashed (37 tests, same 3 failures).

## Honest limits

`robots.txt` is **voluntary and advisory**. It only takes effect on a crawler's next fetch of it, well-behaved crawlers honor it and badly-behaved ones ignore it outright, and `Crawl-delay` is a non-standard extension with uneven support. This is mitigation, not enforcement — a hard cap means rate-limiting at the edge (Cloudflare) or `mod_qos` in front of the application, which is a separate and larger change. This PR is deliberately the small, reversible step.

https://claude.ai/code/session_01ANtyDhUBYCsvEytDfMXUJe


rdhyee:
Bumping this: **another crawler-saturation episode hit prod on 2026-09-02** (~17 min of 503s, self-recovered — see #1172). ClaudeBot and GPTBot were the top three clients by request count during the window.

Checked the live `https://unglue.it/robots.txt` today — it disallows a handful of paths for `User-agent: *` and has **no `Crawl-delay` and no ClaudeBot/GPTBot-specific rules at all**. So these crawlers are currently entirely unthrottled, which is what this PR was written to address.

This is now the recommended first mitigation: smallest change, highest leverage, and both bots do honor robots.txt. Worth noting the limits honestly — robots.txt is voluntary and only takes effect on a crawler's next fetch, so it's mitigation rather than enforcement; a hard cap would mean rate-limiting at the edge or `mod_qos`. But that's a much bigger change, and this one has been sitting since June.

rdhyee:
**[Codex via CC] Round 5 (final) verdict: no blocking defects found.**

Five adversarial rounds. Every round but the last found something real, and the pattern is worth recording: **four consecutive rounds each caught a false factual claim about a specific crawler**, not a code defect. The rules behaved correctly throughout; the stated *reasons* kept being wrong.

| Round | Finding | Resolution |
|---|---|---|
| R1 | `PerplexityBot` grouped with training crawlers | Verified against Perplexity's docs — it's their **search** crawler, "not used to crawl content for AI foundation models." Removed. |
| R2 | `Google-Extended` costs Gemini grounding | Followed through: it and `Applebot-Extended` aren't crawlers at all, so they shed **zero** load. Removed as out of scope. Also: tests asserted `Crawl-delay` but none of the expensive-path exclusions, and 3 of 9 blocked agents. Made exhaustive. RFC wording corrected. |
| R3 | Tests weren't database-isolated | Real defect. The request loads the root URLconf, which imports a module that queries at import time — under `SimpleTestCase` that reached the configured database. Switched to `TestCase`. Also: CCBot's rationale was false (Common Crawl **does** honor `Crawl-delay`); `cohere-ai` is not a documented agent; parser split groups on extension records. |
| R4 | `Diffbot` also documents `Crawl-delay` | Fourth per-crawler claim to prove wrong, so removed the class of claim rather than patching one more instance. The comment now states policy and tells the next editor to check current documentation. |
| R5 | Guard ignored `Allow` rules; two comments overstated | Guard now records `Allow` and rejects any allow beneath a baseline prefix (RFC 9309 longest-match). Comments corrected. |

**One R5 nit declined deliberately:** raw `HTTP_HOST` comparison means `unglue.it:443` and `Unglue.it` fall to the non-production branch and serve `Disallow: /`. Real, but pre-existing gate behavior, orthogonal to crawler policy, and costly to get wrong in either direction — it deserves its own change and its own test. Same applies to `www.unglue.it`.

**Verification.** Every guard was checked by breaking the file on purpose and confirming failure: reverting the pre-fix template trips the precedence guard; deleting `/api/` or the `Bytespider` stanza trips the exhaustive assertions; injecting `Allow: /accounts/public/` trips the new override check. `RobotsTxtTests` 4/4 pass. Full `frontend` suite 41 tests / 3 failures — the same three pre-existing `RhPageTests`, confirmed by running with this change stashed.

**Standing caveat for the merge decision:** `robots.txt` is advisory. Well-behaved crawlers honor it, badly-behaved ones don't, and it only takes effect on their next fetch. A hard cap is edge rate-limiting (Cloudflare) or `mod_qos` — separate and larger.

🤖 Posted by Claude Code at Raymond's direction.

Issue #1142, opened 2026-05-06
Stripe integration: due diligence, modernization arc, and tracking


rdhyee:
## Purpose

Umbrella tracking issue for the Stripe integration in regluit:
- Capture the current state and architecture
- Track the open due-diligence questions raised at the 2026-04-16 and 2026-04-30 Eric meetings
- Track the long-term modernization arc (API version, PaymentIntent, webhook signing)
- Provide a single place to link related issues and PRs

This is **not** a single-deliverable issue. It will stay open across the Django 4.2 migration (PR #1123) and beyond.

Knowledge map (living): kept in Raymond's notes; this issue is the public anchor.

## Context

Eric flagged at the 2026-04-30 meeting that Stripe DD is the named gate for the Django 4.2 migration ship. Doing the DD revealed that we don't have a single place that documents how Stripe is wired into regluit, what's on the modernization arc, or what's known to be broken (#1125). This issue fills that gap.

## Architecture, in one paragraph

regluit's \`payment/\` app abstracts payment processors via \`baseprocessor.py\`. Today only \`stripelib.Processor\` is live (PayPal Adaptive and Amazon Payments are vestigial). The \`Transaction\` model in \`payment/models.py\` is the canonical record (with \`host='stripelib'\` for modern rows); \`PaymentResponse\` is the raw audit log; \`Account\` links a regluit user to a Stripe Customer. \`manager.py\` handles transaction lifecycle and dispatches to \`stripelib.py\`, which calls the Stripe API. \`views.py\` hosts the webhook receiver and donation endpoints. Three campaign types (REWARDS, B2U, THANKS) interact with Stripe differently; THANKS is by far the dominant active flow.

## Yellow flags (known)

- **Stripe API version pinned to \`2013-02-13\`** in \`stripelib.py\`. Stripe is backwards-compat-friendly but this is 12 years old. Modernization to \`PaymentIntent\` / \`PaymentMethod\` is a bounded project, not a Django 4.2 prerequisite.
- **Webhook signature not verified**. The \`stripe_webhook\` view doesn't validate \`Stripe-Signature\`. Bounded risk based on what we do with events, but should be fixed.
- **Anonymous Stripe Customer for logged-in donors** — see #1125.
- **\`Receiver\` table is dead** — vestige of PayPal Adaptive Payments era. Dead code complicates the payment paths.

## Due-diligence checklist (Phase B of #1123)

[...]

PR #1174, opened 2026-06-18
Fix #1125: link logged-in donors to their Stripe Customer (not 'anonymous user')

### LINKED ISSUE #1125: Logged-in donations create anonymous Stripe Customer due to token-presence check
## Bug

When a logged-in user makes a donation on unglue.it, the resulting Stripe Customer is created with `description='anonymous user'` instead of the user's username and email. The charge goes through successfully, but the Stripe dashboard shows no identifying information for the customer.

## Root Cause

In [`payment/stripelib.py`](https://github.com/Gluejar/regluit/blob/master/payment/stripelib.py), the `Pay.__init__` method uses the presence of a `token` parameter as a proxy for "user is anonymous":

```python
if token:
    # user is anonymous
    account = transaction.get_payment_class().make_account(
        token=token, email=transaction.receipt)
```

This calls `make_account` **without passing `user`**, which hits the anonymous branch:

```python
def make_account(self, user=None, token=None, email=None):
    if user:
        customer = sc.create_customer(card=token, description=user.username,
                                      email=user.email)
    else:
        customer = sc.create_customer(card=token, description='anonymous user', email=email)
```

But a logged-in user entering card details for a donation **also** has a token (the Stripe.js token from the card form). The donation path in `FundView.form_valid` correctly sets `self.transaction.user = self.request.user`, but then passes `token=stripe_token` to `p.charge()`, triggering the anonymous code path.

## Steps to Reproduce

1. Log in to unglue.it
2. Make a donation (any amount)
3. Enter card details
4. Check Stripe dashboard — the Customer shows as "anonymous user"

## Expected Behavior

The Stripe Customer should be created with the logged-in user's username and email. The fix should check `transaction.user` (not just token presence) when deciding whether to pass user identity to `make_account`.

## Additional Context

- The Stripe API version in the code is `'2013-02-13'` — 13 years old
- The `create_customer` call never passes a `name` parameter (only `description` and `email`)
- The `create_charge` call passes no `receipt_email`, `metadata`, or `statement_descriptor`
- `Transaction.anonymous` is a separate site-level display preference and is unrelated to this bug
- Confirmed by a $5 test donation on 2026-04-16

## Related

- PR #1123 (Django 4.2 upgrade) includes Stripe 2.76 → 5.5 bump — this bug should be considered as part of payment code review
- The fix is small but touches the payment flow, so it should be tested with Stripe test-mode charges

rdhyee:
Fixes #1125.

## The one-line fix
In `payment/stripelib.py`, `Pay.__init__` used **token-presence** as a proxy for "anonymous user." But a logged-in donor entering card details also produces a token, so they were routed through `make_account()`'s anonymous branch → Stripe `Customer description='anonymous user'` and `Account.user=None` (orphaned, no saved-card reuse).

**Fix:** pass `user=transaction.user` into `make_account` (it's `None` for a truly anonymous donor, and the user for a logged-in one). This makes the donation path consistent with the **other `make_account(user=...)` call sites** already in `FundView` (e.g. the `transaction.user.id != request.user.id` branch).

## Behavior note (for review)
`make_account` with `user` set runs the existing `if user and user.profile.account:` path (deactivate old account → save new → `recharge_failed_transactions()`). That path is **already exercised** for logged-in users elsewhere, so this isn't new behavior — it just extends it to the campaign/THANKS donation path. Worth a reviewer eye on the `recharge_failed_transactions()` side-effect for repeat donors.

## Verification
- Bug reproduced on **live** txns **16674** (2026-06-18, the cutover-validation payment) and **15493** (2026-04-16) — both showed "anonymous user" for a logged-in donor.
- ⚠️ Not yet run against Stripe **test-mode** — should be charge-tested before merge (small change, but payment flow).

## Scope
Minimal targeted fix. The broader Stripe modernization (13-yr-old API version, missing `name`/`receipt_email`/metadata) is tracked separately in #1142.

---
*Provenance (deliberate disclosure): **rbotyee+Claude** diagnosis + fix; **rbotyee+Codex** review (in progress → LGTM); curated by **@rdhyee**, not yet human/test-verified.*

### DIFF
--- a/frontend/tests.py
+++ b/frontend/tests.py
@@ -226,8 +226,8 @@ from unittest.mock import patch as _patch
 from django.test import TestCase as _TestCase, Client as _Client
 from django.urls import reverse as _reverse
 from django.contrib.auth.models import User as _User
-from payment.models import Transaction as _Transaction
-from payment.parameters import PAYMENT_HOST_NONE as _PAYMENT_HOST_NONE
+from regluit.payment.models import Transaction as _Transaction
+from regluit.payment.parameters import PAYMENT_HOST_NONE as _PAYMENT_HOST_NONE
 
 
 class FundViewOwnershipTest(_TestCase):
@@ -244,13 +244,13 @@ class FundViewOwnershipTest(_TestCase):
         self.txn = _Transaction.create(amount=5.00, max_amount=5.00,
                                        host=_PAYMENT_HOST_NONE, user=self.owner, campaign=None)
 
-    @_patch('frontend.views.PaymentManager')
+    @_patch('regluit.frontend.views.PaymentManager')
     def test_anonymous_cannot_pay_owned_transaction(self, mock_pm):
         r = _Client().post(_reverse('fund', args=[self.txn.id]), {'stripe_token': 'tok_test'})
         self.assertTemplateUsed(r, 'pledge_user_error.html')
         mock_pm.return_value.charge.assert_not_called()
 
-    @_patch('frontend.views.PaymentManager')
+    @_patch('regluit.frontend.views.PaymentManager')
     def test_other_user_cannot_pay_owned_transaction(self, mock_pm):
         c = _Client()
         c.login(username='other1125', password='pw')
--- a/frontend/tests.py
+++ b/frontend/tests.py
@@ -220,3 +220,41 @@ class GoogleBooksTest(TestCase):
         work_url = r['location']
         self.assertTrue(re.match(r'.*/work/\d+/$', work_url))
 
+
+# --- #1125 ownership regression (authored 2026-06-18; verify locally) ---------
+from unittest.mock import patch as _patch
+from django.test import TestCase as _TestCase, Client as _Client
+from django.urls import reverse as _reverse
+from django.contrib.auth.models import User as _User
+from payment.models import Transaction as _Transaction
+from payment.parameters import PAYMENT_HOST_NONE as _PAYMENT_HOST_NONE
+
+
+class FundViewOwnershipTest(_TestCase):
+    """Regression for #1125 hardening: a no-campaign transaction owned by a user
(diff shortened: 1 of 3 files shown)


rdhyee:
### Codex collaboration → converged (rbotyee+Claude × rbotyee+Codex)

3 review rounds, each caught something real:
1. **R1:** passing `user=transaction.user` let an **anonymous** request mutate an owner's Stripe account → added ownership guard.
2. **R2:** an **authenticated different user** could still take over the txn → **consolidated guard** rejects both; added `FundViewOwnershipTest`.
3. **R3:** guard confirmed correct (closes both vectors; owner / anonymous-unowned paths still proceed). Test import + `@patch` targets needed the `regluit.` prefix → **fixed** (verified against repo convention).

**Status:** guard logic Codex-LGTM'd; tests corrected per repo convention but **not yet executed** (no Django test env here). **Will run `manage.py test frontend` on the rebuilt test.unglue.it (4.2) before marking ready.**

rdhyee:
### ✅ Tests PASS on real Django 4.2 (test.unglue.it)

Ran on the freshly-rebuilt **test.unglue.it** (Django 4.2.21, py3.12, prod-snapshot DB) — the regression test the Codex loop produced now executes green:

```
test_anonymous_cannot_pay_owned_transaction ... ok
test_other_user_cannot_pay_owned_transaction ... ok
Ran 2 tests in 2.319s
OK
```

Test DB built + migrated cleanly; both ownership-guard cases confirmed (anonymous + authenticated-mismatch rejected before `charge()`). `manage.py check` surfaced only pre-existing warnings (django-ckeditor CVE notice → tracked in #1142; MySQL strict-mode hint). **Moving out of draft — guard is Codex-LGTM'd and the regression suite is green on real 4.2.**

rdhyee:
Closing — superseded by #1247, which fixes the same bug (#1125) and was merged on 2026-09-10.

Issue #1081, opened 2026-02-11
Migration plan: strip features, then upgrade Django 1.11 → 5.1


rdhyee:
## Overview

Strip deprecated campaign types first (on Django 1.11 where tests work), then migrate a smaller codebase through Django versions.

**Keep:** THANKS campaigns (type=3) + Donations (transaction.campaign=None)
**Remove:** REWARDS (type=1), BUY2UNGLUE (type=2), Booxtream integration

---

## Phase 0: Bot protection (independent track, see #1078)

- Signed/expiring S3 URLs
- Bot checks on download views
- Can proceed in parallel with everything below

## Phase 1: Strip REWARDS (on Django 1.11)

Most entangled type (Premium model, deadline logic, pledge flow). Remove first.

- Remove `Premium` model, `deadline` field/countdown logic
- Remove REWARDS branch in `handle_transaction_charged()` signal
- Remove REWARDS conditionals in views, forms, templates
- Validate: test suite passes, THANKS + donations still work

## Phase 2: Strip BUY2UNGLUE + Booxtream (on Django 1.11)

Tightly coupled — remove together.

- Remove `Offer` model, `cc_date_initial`, `do_watermark` flag
- Remove `booxtream/` app, `Acq.watermarked` FK
- Remove BUY2UNGLUE payment signal branch, `download_campaign()` view, `watermark_acq` task
- Keep: `Acq` model (THANKS uses it), `ask_pdf()`/`ask_epub()` personalization
- Validate: THANKS + donations still work, download flow simplified

## Phase 3: Cleanup (still Django 1.11)

- Collapse dead conditional branches (70+ in Python, 85+ in templates)
- Simplify `manage_campaign.html` (currently 24+ type checks)
- Simplify payment forms
- DB migration to drop removed columns/tables
- Full test suite + manual smoke test on staging

## Phase 4: Django upgrade

Smaller codebase, standard upgrade path:

```
1.11.29 → 2.2 LTS → 3.2 LTS → 4.2 LTS → 5.1
```

## Phase 5: Post-migration modernization

- Python 3.8 → 3.12+
- Cloudflare proxy (full WAF/bot management)
- Other infrastructure improvements

---

## Dependencies

```
Phase 0 (bots) ─────────────────────────────────> independent
Phase 1 (REWARDS) → Phase 2 (B2U) → Phase 3 (cleanup) → Phase 4 (Django) → Phase 5
```

## Entanglement analysis (why strip first)

[...]

PR #1202, opened 2026-07-03
Django 5.2 prep: STORAGES dict + drop USE_L10N (#1081)

rdhyee:
First concrete step toward the Django 4.2 → 5.2 LTS upgrade ([#1081](https://github.com/Gluejar/regluit/issues/1081)). **Forward-compatible** settings changes that behave identically on the current 4.2 runtime and remove two 5.x landmines.

## Changes
- **`STORAGES` dict** replaces `DEFAULT_FILE_STORAGE` (removed in Django 5.1). `STORAGES` is supported since **4.2**, so this is a no-op on the live runtime; the S3 (`django-storages`) vs `FileSystemStorage` selection is preserved.
- **Drop `USE_L10N`** (removed in Django 5.0; localized formatting is always on).

## Why it's small
The other landmines from `PLAN_django52_upgrade.md` are already clean here: no `django.utils.timezone.utc`, no `index_together`, no removed util imports (`force_text`/`ugettext`/…). The real remaining work is the dependency audit (`django-selectable`, `django-registration`, ckeditor5) — separate PRs.

## ⚠️ Validation needed before merge
The `STORAGES` change touches the **prod S3 media path**. Confirm media upload + serve on **test.unglue.it** before merging (per the plan's "test, don't just rename").

No code reads `settings.DEFAULT_FILE_STORAGE` directly (checked), and no `STATICFILES_STORAGE` was set, so static handling is unchanged.

### DIFF
--- a/settings/common.py
+++ b/settings/common.py
@@ -41,9 +41,8 @@ SITE_ID = 1
 # to load the internationalization machinery.
 USE_I18N = True
 
-# If you set this to False, Django will not format dates, numbers and
-# calendars according to the current locale
-USE_L10N = True
+# USE_L10N was removed in Django 5.0 (localized formatting is always on), so it
+# is intentionally not set here.
 
 # Absolute filesystem path to the directory that will hold user-uploaded files.
 # Example: "/home/media/media.lawrence.com/media/"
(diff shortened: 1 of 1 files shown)


rdhyee:
🤖 **Codex review: LGTM** (no changes). Confirmed: `STORAGES` is behavior-identical on Django 4.2 (same S3-vs-FileSystem selection, just relocated); explicit `staticfiles` entry matches Django's default; `DEFAULT_FILE_STORAGE` removed so no `ImproperlyConfigured` risk; `USE_L10N` removal is a no-op (default already `True`). Reminder: still validate S3 media upload/serve on test.unglue.it before merge.

rdhyee:
⚠️ **Deploy coordination required** (found by Codex during the #1203 spike review): the Ansible-rendered `prod.py` (provisioning `prod.py.j2:175`) still sets `DEFAULT_FILE_STORAGE` after common.py — on Django 4.2 that combination raises `ImproperlyConfigured` (reproduced locally) → **site-wide 500 at boot if this PR deploys alone.** Merge+deploy EbookFoundation/regluit-provisioning#61 in the same window (config render before/with app restart).

PR #1203, opened 2026-07-03
Django 4.2 → 5.2 LTS upgrade (#1081)

rdhyee:
[#1081](https://github.com/Gluejar/regluit/issues/1081) — the Django 5.2 LTS upgrade, **rebased 2026-08-23 onto post-comments-removal master** (`4a0a51b0`). Six files, +22/−9. Eric green-lit reviewing this 2026-08-10; **merge to master still gated on staging soak + Eric's look.**

## What changed in the rebase (vs the original spike you may have read)

- **`django-contrib-comments` bump is GONE** — the app was removed entirely by #1217/#1220, both live on production 2026-08-23. The `django_comments.0004` migration vanishes from this upgrade's path. That was the point of sequencing comments first.
- The #1202 STORAGES/USE_L10N prep is already on master — no longer a co-requisite.
- Everything else is the reviewed spike verbatim (confirmed by range-diff in Codex's re-review).

## The two judgment calls, unchanged

- **`USE_TZ = False` pinned**: Django 5.0 flipped the default; this codebase and its data are naive-datetime throughout. TZ-awareness is a separate project.
- **`Campaign.clone()` reorder**: 5.x raises on filtering related objects of unsaved instances. On the post-#1213 base this is lower-risk still — only THANKS campaigns are clonable, and THANKS `update_left()` sets `left=0` without querying transactions.

## Verification (2026-08-23, local MySQL 8, both spike envs on the identical tree)

- core + 16 scrub tests under 5.2.15: failure set **identical** to the 4.2 baseline (known offline/live-API failures only, zero new)
- frontend / payment / libraryauth: no new failures. (Two libraryauth errors that vary between versions in full-suite order are a pre-existing import-order fragility in `puzzle.py` cache snapshotting — they error under **both** versions in isolation; not upgrade evidence.)
- `migrate --plan` against a current-production-schema DB: **no pending migrations**
- Codex re-review of the rebased diff (2026-08-23): **LGTM for staging**; production waits for soak + Eric

## Soak plan

Staging (test.unglue.it = exact prod copy, same code+data shape as production as of today) gets this branch with `run_pip=true run_collectstatic=true`, then: Django-version-in-venv check, cold-start registration path, THANKS-only campaign smoke tests, historical-notification rendering, sass/storage pages, and log watch for naive/aware datetime and pickle errors. Prod deploy only after a clean multi-day soak and Eric's review.

### DIFF
--- a/booxtream/tests.py
+++ b/booxtream/tests.py
@@ -61,7 +61,7 @@ class TestBooXtream(unittest.TestCase):
             }
         params['referenceid'] = 'order' + str(time.time())
         boox = inst.platform(epubfile=self.epub2file, **params)
-        self.assertRegexpMatches(boox.download_link_epub, 'download.booxtream.com/')
+        self.assertRegex(boox.download_link_epub, 'download.booxtream.com/')
         self.assertFalse(boox.expired)
         self.assertEqual(boox.downloads_remaining, 3)
 
@@ -71,4 +71,4 @@ class TestBooXtream(unittest.TestCase):
         in_mem_epub.write(self.epub2file.read())
         in_mem_epub.seek(0)
         boox2 = inst.platform(epubfile=in_mem_epub, **params)
-        self.assertRegexpMatches(boox2.download_link_epub, 'download.booxtream.com/')
+        self.assertRegex(boox2.download_link_epub, 'download.booxtream.com/')
--- a/core/models/__init__.py
+++ b/core/models/__init__.py
@@ -422,8 +422,12 @@ class Campaign(models.Model):
             self.created = None
             self.name = 'copy of %s' % self.name
             self.activated = None
-            self.update_left()
+            # save() BEFORE update_left(): update_left() -> current_total filters
+            # Transactions on this instance, and Django 5.x raises ValueError for
+            # unsaved instances in related filters (4.2 silently matched nothing).
+            # A fresh clone has no transactions, so the computed totals are identical.
             self.save()
+            self.update_left()
             self.managers.set(old_managers)
 
             # clone associated premiums
--- a/core/tests.py
+++ b/core/tests.py
@@ -1099,8 +1099,8 @@ class EbookFileTests(TestCase):
         self.assertIsNot(acq.nonce, None)
 
         url = acq.get_watermarked().download_link_epub
-        self.assertRegexpMatches(url, 'github.com/eshellman/42_ebook/blob/master/download/42')
-        #self.assertRegexpMatches(url, 'booxtream.com/')
+        self.assertRegex(url, 'github.com/eshellman/42_ebook/blob/master/download/42')
+        #self.assertRegex(url, 'booxtream.com/')
 
         with self.assertRaises(UnglueitError) as cm:
             c.activate()
--- a/libraryauth/tests.py
+++ b/libraryauth/tests.py
@@ -88,7 +88,9 @@ class TestAppConfigSignalsWired(TestCase):
         # ready() must have imported signals.py and connected the dedup receiver.
         from django_registration.signals import user_activated
         names = []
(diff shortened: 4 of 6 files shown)


rdhyee:
**Codex review: LGTM** (no high-confidence blockers). Notable points, all captured:
- `USE_TZ=False` pin confirmed correct and necessary.
- `Campaign.clone()` reorder: final persisted state equivalent; only a transient pre-`update_left()` row visible to concurrent readers/DB triggers (none in-repo). Non-blocking.
- Dep bumps: recommend a `collectstatic`/SCSS smoke on staging (sass-processor 1.x changed `--use-storage` behavior; repo doesn't use the removed names). comments 2.2 migrations touch real comment data (an index) — fine but real.
- **Real catch: the rendered prod `prod.py` still sets `DEFAULT_FILE_STORAGE`** → mutually-exclusive error on 4.2 the moment #1202 deploys (reproduced locally). Fix filed: EbookFoundation/regluit-provisioning#61 — **must deploy in the same window as #1202.**

rdhyee:
**[CC]** Rebased onto current `master` and taken out of draft (2026-08-17).

**What changed:** the branch was 16 commits behind `master` (branch point `e82e5996`, 2026-07-03). Rebased `a3e9cf86` → `9e3e12c2`. **One conflict**, in `requirements.txt`: `master` had added `nh3` adjacent to the lines this PR bumps. Resolved by keeping both. No other file conflicted — the `STORAGES` / `USE_L10N` hunks auto-merged because [#1202](https://github.com/Gluejar/regluit/pull/1202) landed them on `master` on 2026-07-22.

**Net against current master: 6 files, +23/−10.**

## Two corrections to this PR's description

**1. "Land #1202 first" is already satisfied** — #1202 merged 2026-07-22; `master` carries the `STORAGES` dict and no longer sets `USE_L10N`. That step can be struck.

**2. The `Campaign.clone()` rationale understates the fix.** The description says Django ≤4.2 gave a "silent no-match". It doesn't. 4.2 normalizes the unsaved instance to `pk=None` and renders `filter(campaign=self)` as `campaign_id IS NULL` — and `Transaction.campaign` is `null=True` (`payment/models.py:115`). So on 4.2 that query could match **uncampaigned transactions** (donations and similar), fold them into `current_total`, and produce a wrong `left` on a freshly cloned campaign.

Whether qualifying rows exist is empirical against production data, so this is *plausibly corrects a latent data bug*, not a proven one — but it's a point in the upgrade's favour and the "matched nothing" wording is worth not repeating.

Minor precision: the reorder isn't strictly behaviour-preserving either. The end state of a successful clone is identical, but the first INSERT now carries the source campaign's copied `left` until the subsequent `update_left()` UPDATE corrects it, with no atomic wrapper around the pair. Low impact — cloning is rare, and post-#1213 only THANKS campaigns are clonable (`left = 0`).

## ⛔ This will need ONE more rebase before merge

Agreed order is #1213 → #1217 → #1218/#1220 → **this PR**. [#1217](https://github.com/Gluejar/regluit/pull/1217) **deletes** `django-contrib-comments` from `requirements.txt` and `Pipfile` entirely, so the `2.0.0 → 2.2.0` bump carried here becomes a modify/delete conflict. **Resolution at that point: drop the bump.** It is kept for now only because current `master` still has the package.

That coupling is the concrete reason for the ordering: doing 5.2 first would mean carefully upgrading a package that is about to be deleted.

## Still outstanding before merge

- Deploy to test.unglue.it **with `-e run_pip=true`** (dependencies changed) and run the smoke gate
- Soak, then release per the runbook
- `USE_TZ = False` is a deliberate deferral, now tracked as [#1221](https://github.com/Gluejar/regluit/issues/1221) rather than living only in a settings comment

PR #1209, opened 2026-07-15
Stop feedback page linking to itself (crawler-trap root cause)

rdhyee:
**Why:** removes the app-level root cause of the 2026-07-10 crawler-trap flood — the mirror-image companion to the Apache guard in [EbookFoundation/regluit-provisioning#63](https://github.com/EbookFoundation/regluit-provisioning/pull/63) (see also EbookFoundation/regluit-provisioning#65). Incident: `INCIDENT_2026-07-10_crawler_trap_flood.md`.

## The hall-of-mirrors problem (plain English)

Every page's footer and nav emitted a "Feedback" link of the form:

```
{% url 'feedback' %}?page={{request.build_absolute_uri|urlencode:""}}
```

i.e. `/feedback/?page=<the-current-page-url>`, so the feedback form can record where the user came from. Harmless on most pages — but that footer renders on the `/feedback/` page **too**, so the feedback page linked to a feedback-page-about-the-feedback-page, which linked to a feedback-page-about-that, forever. Each hop re-URL-encodes the last (`%2F` → `%252F` → `%25252F`), so it's an infinite, ever-growing URL space. On 2026-07-10 datacenter bot fleets walked it at ~60k requests/hour and pegged all four web workers at ~90% CPU (page loads went from ~1s to 17–31s). `/feedback/` GETs alone were ~7% of traffic (84k on 7/9).

The provisioning hotfix (#63) now serves cheap Apache 403s for the recursive requests, but the app should never emit the self-referencing link in the first place. This PR does that.

## What changed

- New template tag `frontend/templatetags/feedback_link.py` exposing `{% feedback_url %}`:
  - On the feedback page itself (detected via `request.resolver_match.url_name == 'feedback'`, falling back to `request.path == reverse('feedback')`) it returns a bare `{% url 'feedback' %}` — no `?page=`, breaking the loop.
  - On every other page it appends `?page=<urlencoded current absolute uri>`, exactly as before (`quote(..., safe='')` matches the old `urlencode:""`).
  - It also strips any pre-existing `page=` query param from the current URL before embedding it, so other query strings can't reintroduce a level of nesting.
  - If there's no `request` in context, it returns the bare URL.
- Replaced the inline pattern with `{% feedback_url %}` (and added `{% load feedback_link %}`) in the **12 crawler-reachable HTML page templates** that carried it: `base.html` (×2, the footer/nav that caused the trap), `add_your_books.html`, `download.html`, `faq.html`, `gift_error.html`, `gift_user_error.html`, `gift_welcome.html`, `manage_campaign.html`, `pledge_card_error.html`, `pledge_user_error.html`, `rh_tools.html`, `supporter.html`.
- **Intentionally not changed:** the 6 `notification/*.html` email templates that use the same pattern. They render without a live request (crawlers don't fetch emails), so they aren't part of the trap; leaving them keeps the change scoped to the crawl surface.

## Feedback view is safe when `page` is absent

`feedback()` in `frontend/views/__init__.py` already does `try: context['page'] = request.GET['page]` / `except: context['page'] = '/'`, so a bare `/feedback/` (no param) has always worked and is unaffected.

## How verified

- **Tag logic, rendered standalone** (minimal Django template engine + stub URLconf, no DB): confirmed all four cases — non-feedback page carries the fully-encoded `?page=`; bare `/feedback/` and `/feedback/?page=...` both emit plain `/feedback/`; a page with its own `page=` param has it stripped before embedding; missing request → bare URL. Non-feedback output byte-for-byte matches the old `urlencode:""` encoding.
- **Regression tests added** to `frontend/tests.py` (`FeedbackSelfLinkTests`): `/feedback/` HTML must not contain `/feedback/?page=`; `/privacy/` (a fixture-free page) must still contain it. These follow the repo's existing `TestCase`+`Client` pattern.
- **Not run locally:** the full Django test suite (needs a test DB + vault-decrypted settings) was not executed in this environment. The added tests are provided for CI / a DB-backed run.

https://claude.ai/code/session_01CZoSkYdkKLH37tH4QbEmH7

### DIFF
--- a/frontend/templates/add_your_books.html
+++ b/frontend/templates/add_your_books.html
@@ -1,4 +1,5 @@
 {% load cf %}
+{% load feedback_link %}
 	<script type="text/javascript" src="/static/js/watermark_init.js"></script>
 	<script type="text/javascript" src="/static/js/watermark_change.js"></script>
 	<p id="add_your_books"><b>Claiming a work</b></p>
@@ -20,5 +21,5 @@
 	<ul class="bullets">
 	    <li>Use <a href="{% url 'new_edition' '' '' %}">this form</a> to enter the metadata.</li>
 	    <li>Your metadata should have title, authors, language, description.</li>
-	    <li>If you have a lot of books to enter, <a href="{% url 'feedback' %}?page={{request.build_absolute_uri|urlencode:""}}">contact us</a> to load ONIX or CSV files for you.</li>
+	    <li>If you have a lot of books to enter, <a href="{% feedback_url %}">contact us</a> to load ONIX or CSV files for you.</li>
 	</ul>
--- a/frontend/templates/base.html
+++ b/frontend/templates/base.html
@@ -1,5 +1,6 @@
 <!DOCTYPE html>
 {% load sass_tags %}
+{% load feedback_link %}
 
 <html>
 <head>
@@ -39,7 +40,7 @@
 
 <body>
 <div id="feedback">
-<p><a href="{% url 'feedback' %}?page={{request.build_absolute_uri|urlencode:""}}" class="nounderline">Feedback</a></p>
+<p><a href="{% feedback_url %}" class="nounderline">Feedback</a></p>
 </div>
 
 <div id="about_expandable">
@@ -163,7 +164,7 @@
             <li><a href="{% url 'faq' %}">General FAQ</a></li>
             <li><a href="{% url 'faq_location' 'rightsholders' %}">Author/Publisher FAQ</a></li>
             <li><a href="{% url 'api_help' %}">API</a></li>
-            <li><a href="{% url 'feedback' %}?page={{request.build_absolute_uri|urlencode:""}}">Support</a>
+            <li><a href="{% feedback_url %}">Support</a>
             <li><a href="{% url 'libraries' %}">Unglue.it for Libraries</a>
         </ul>
     </div>
--- a/frontend/templates/base.html
+++ b/frontend/templates/base.html
@@ -109,10 +109,10 @@
             {% else %}
                 <div class="js-topmenu">
                     <ul class="menu">
-                        <li><a class="notbutton hijax" href="{% url 'superlogin' %}?next={% if request.GET.next %}{{ request.GET.next|urlencode }}{% else %}{{ request.get_full_path|urlencode}}{% endif %}"><span>Sign In</span></a></li>
+                        <li><a class="notbutton hijax" href="{% url 'superlogin' %}?next={% auth_next %}"><span>Sign In</span></a></li>
                        {% if not suppress_search_box %}
                             {% if request.get_full_path != "/accounts/register/" %}
-                                <li  class="last"><a class="btn btn-signup" href="{% url 'registration_register' %}?next={% if request.GET.next %}{{ request.GET.next|urlencode }}{% else %}{{ request.get_full_path|urlencode}}{% endif %}">Sign Up <i class="fa fa-chevron-right"></i></a></li>
+                                <li  class="last"><a class="btn btn-signup" href="{% url 'registration_register' %}?next={% auth_next %}">Sign Up <i class="fa fa-chevron-right"></i></a></li>
                             {% endif %}
                         {% endif %}
                     </ul>
--- a/frontend/templates/download.html
+++ b/frontend/templates/download.html
@@ -1,4 +1,5 @@
 {% extends 'base.html' %}
+{% load feedback_link %}
 
 {% load humanize %}
 {% load sass_tags %}
@@ -392,7 +393,7 @@ $j(document).ready(function() {
         
         <br />
         <h4>Need more help?</h4>
-        <p><a href="{% url 'feedback' %}?page={{request.build_absolute_uri|urlencode:""}}">Ask us a question!</a></p>
+        <p><a href="{% feedback_url %}">Ask us a question!</a></p>
         
         <div>
             <a href="http://www.defectivebydesign.org/drm-free"><img src="/static/images/DRM-free150.png" alt="DefectiveByDesign.org" border="0" align="middle" /></a> All downloads from Unglue.it are DRM free. Hooray!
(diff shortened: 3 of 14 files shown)


rdhyee:
**Codex review (gpt-5.4, 2026-07-15): request changes.** Full findings applied in the follow-up commit:

1. **Medium — pagination regression**: `_strip_page_param()` removed `page=` from every recorded URL, breaking legitimate pagination context (e.g. feedback from `/search/?q=…&page=2`). The feedback-route guard alone terminates the recursive chain (depth can never exceed 1), so stripping is unnecessary → removed; non-feedback pages now record the exact `build_absolute_uri()`.
2. **Medium — unbounded `parse_qsl()`** DoS surface → moot after removing the stripping.
3. **Medium — tests too narrow** → added: pagination-preservation test, exact-value test, feedback-page-with-page-param test, missing-request handling.
4. **Low, pre-existing**: `FeedbackForm.page` accepts arbitrary URL schemes (`javascript:`) and re-emits them as the post-submit 'go back' href. Predates this PR — filed separately.

What Codex confirmed correct: tag idiom, escaping, resolver_match/None handling, `quote(page, safe='')` equivalence with the original `urlencode:""` filter, IRI handling via `build_absolute_uri()`.

rdhyee:
**Codex review round 3 (gpt-5.4, 2026-07-15): LGTM.**

> No remaining correctness, security, performance, or maintainability blockers found. The chain is bounded: /feedback/ always produces `next=%2Ffeedback%2F`; both Sign In and Sign Up use auth_next; superlogin/registration reuse the incoming next rather than wrapping the current URL. Executable-modeled five rounds on both routes — auth and feedback URLs stayed at fixed lengths. Fully percent-encoded (`safe=''`), no new injection surface, no meaningful performance regression.

Review history: round 1 caught a pagination regression (fixed in `a5833953`); round 2 caught the feedback↔login next-param growth chain (fixed in `4a762f24`); round 3 clean. Non-blocking suggestion noted for a follow-up: parameterize the fixed-point test to also walk registration_register explicitly.

CC review: diff read in full at each round; both Codex findings verified before fixing. Not run locally: the Django test suite (pinned py3.9 env unavailable) — the 8 new/updated tests should be exercised in CI or on test.unglue.it before deploy.

rdhyee:
## BLUF — what this PR fixes, with log evidence from the 2026-07-10 incident

**Bottom line: unglue.it's own templates created an unbounded, self-generating URL space; a single dumb scraper walked it for 25,486 requests in one day (17,337 *distinct* URLs — no URL fetched more than twice); this PR makes that URL growth reach a fixed point instead. Test-verified (6/6 new tests green on test.unglue.it) and ready to merge.**

### The evidence (from `20260710_access.log`, IP 37.6.162.165)

| Measure | Value |
|---|---|
| Total requests | 25,486 (one day, one IP) |
| Distinct URLs | 17,337 — max repetition of any single URL: **2** |
| URL length | median **1,118 chars**, max 1,834 |
| First requests (10:11a) | `/work/471838/`, `/work/471838/download/` — then `superlogin/?next=`, `register/?next=`, and into the trap |

One real URL, decoded (each level was live-rendered by Django):

```
/password/reset/?next=
  /feedback/?page=
    https://unglue.it/password/reset/?next=
      /feedback/?page=
        https://unglue.it/password/reset/?next=
          /work/471838/download/
```

Each embedding layer adds a round of percent-encoding (`?` → `%3F` → `%253F` → `%25253F`), so encoding depth literally counts the recursion depth.

### The mechanism

Anonymous pages inheriting `base.html` emit two families of parameterized links: the footer's `/feedback/?page=<current-URL>` and Sign-In/Sign-Up's `?next=<current-path>`. Certain edges in that link graph are **recursive**: the feedback page's own links embed the feedback URL (with its `page=` param) inside a new URL, auth pages' feedback links embed the auth URL, and so on. A crawler that follows those particular edges gets a *fresh, longer* URL on every round — the nesting pattern in the logs is exactly this cycle. (The crawler-generated interpretation is an inference, but the encoding-depth pattern supports no other reading.)

### What the fix changes

- **`{% feedback_url %}`**: on the feedback page itself, the footer feedback link drops `?page=` — the direct self-reference edge is gone.
- **`{% auth_next %}`**: on the feedback route, Sign-In/Sign-Up links use the bare path (preserving an *incoming* `next=` when present, so real user flows are untouched) — the second growth edge found in review round 2.

Together they remove the unbounded growth edges in the feedback/auth/password-reset cycle: following it now revisits **bounded, repeating URLs** instead of adding an encoded layer each round. (Scope note: this bounds *this* cycle; it is not a claim that the site's whole crawlable space is finite — the faceted-search space is a separate, known issue.)

### Defense-in-depth context

The Apache guards from [provisioning#63](https://github.com/EbookFoundation/regluit-provisioning/pull/63) stay in place: 403 on triple-encoded slashes (`%25252F`), on `/feedback/` queries embedding an encoded `/feedback`, and on the scraper's UA. Those are the belt at the door; this PR removes the trap itself.

Related, pre-existing, filed separately: #1210 (`page` param accepts `javascript:` URLs).

*Side-finding from the same investigation ([provisioning#64](https://github.com/EbookFoundation/regluit-provisioning/issues/64#issuecomment-4984010358)): the scraper was never login-walled — the book's PDF downloads anonymously (302→S3, verified live 7/15); it wedged purely in this link cycle.*

---
🤖 Writeup drafted by Claude Code from primary log analysis; fact-checked by Codex (gpt-5.4, round 4) — its corrections (scope of the bounded-space claim, `auth_next` next-preservation, exact Apache rule semantics) are incorporated. Posted on Raymond's behalf.

rdhyee:
**Deployed to production** 2026-07-15 ~11:57a PT via release PR #1211 → `production @ 94998ea9` (ansible deploy.yml, per runbook).

Smoke gate green on prod: site 200 (TTFB 0.8s) · `/feedback/` emits bounded links (footer `/feedback/` no `?page=`; auth `next=%2Ffeedback%2F`) · normal pages keep `?page=` · search/work/download/login green · apache2+celeryd+celerybeat active · **0 5xx post-restart** (22 in-flight drops during the apache bounce minute only). Rollback point if needed: `c99e1993`.

Test leg: `master @ 35a6c093` on test.unglue.it, same gate green, 6/6 tests.

🤖 Deployed/verified by Claude Code on Raymond's behalf.

PR #1217, opened 2026-08-06
Remove Django comments (#1130) — phase 1: code removal + queue-safety scrub

rdhyee:
Removes Django comments functionality per [#1130](https://github.com/Gluejar/regluit/issues/1130) — Eric-approved (2026-04-23, re-confirmed 2026-08-03): comments are a spam magnet with low-value content, and removing the app before the Django 5.2 upgrade eliminates the `django_comments.0004` index migration from the 5.2 path entirely.

## Phase 1 of 2 — code removal + one non-destructive migration

This PR removes every runtime reference to `django_comments`; the DB tables and notice-type rows stay behind, orphaned and unreferenced. A follow-up migration-only PR (phase 2, `comments-drop-migration-1130`) deletes the notice types and drops the tables — it merges **only after this release is deployed and all processes restarted**, which makes the destructive step safe under every deploy path (no restart/migrate ordering to get wrong; Codex R2 finding).

The one migration here (`core.0031_scrub_comment_notice_batches`) deletes pending notification-queue batches that carry comment notices. It must ship with this release (Codex R3 HIGH): those batches embed pickled `Comment` instances that the comment-free code can neither unpickle nor render, and `notification.engine.send_all()` blocks on the first failing batch — one stale spam-comment notification would wedge the entire queue. The scrub is non-destructive and safe under any deploy ordering (old code is indifferent to deleted pending batches); phase 2 re-runs it to catch anything created in the migrate→restart window.

## ⛔ Merge order

**Merge after [#1213](https://github.com/Gluejar/regluit/pull/1213).** This branch's migration `core.0031` depends on #1213's `core.0030_retire_pledge_b2u` — the agreed sequencing (#1213 → #1130 → 5.2) is encoded structurally, and `migrate` fails loudly (NodeNotFoundError) if mis-ordered.

## What's removed

- **App wiring**: `django_comments` out of `INSTALLED_APPS`; `/comments/` URL includes (both the django_comments URLs and the site's latest-comments page)
- **Work page**: Comments tab gone. Tab numbering **deliberately stays 1/3/4** — `tabs4.js` and the CSS bind by fixed class names (`tabs2`, `#tabs-2`), and empty jQuery selections no-op, so no JS/CSS churn. Inbound `?tab=2` links (old notification emails) degrade to the Description tab via the view's whitelist.
- **Homepage activity feed**: comment events out of the pledges/wishes/comments chain (feed logic and template branch)
[...]

PR #1213, opened 2026-07-22
Retire Pledge & Buy-to-Unglue campaigns (#1195)

rdhyee:
Closes [#1195](https://github.com/Gluejar/regluit/issues/1195). **Retires Pledge & Buy-to-Unglue**: no new campaigns of those types can be created or launched — not via the form, a forced POST, or `activate()` — every path fails closed with a clear message. The two live legacy campaigns ([#126 Lagos 2060](https://unglue.it/work/128685/), [#137 Contact With Shadow](https://unglue.it/work/137647/)) **keep running unchanged**. ToS updated. Deployed & rehearsed on test.unglue.it; full suite = master baseline (zero new reds); `CC+Codex+LGTM`; Raymond has fine-tooth-read the user-facing text (8/9).

## ⚖️ One choice, then two small calls (Eric)

- [ ] **1. Pick A or B** — same behavior, differs only in how much *legacy text* user-facing pages carry:

| | What readers see | Where to look |
|---|---|---|
| **A — labeled notes** (this PR as-is) | Marketing/help pages keep short "retired programs" notes | [Files tab](https://github.com/Gluejar/regluit/pull/1213/files) |
| **B — strict removal** (A + 3 stacked commits) | Zero legacy text on marketing/help pages; the record lives in the ToS (which links a legacy FAQ at `/faq/campaigns/b2u/`) | [B's extra diff — 5 files, ~20 lines](https://github.com/Gluejar/regluit/compare/retire-pledge-b2u-1195...retire-1195-strict-legacy) |

Raymond's lean: **B**. If B: the stacked commits fast-forward into this branch; nothing else changes.

- [ ] **2. Legal read of the ToS diff** ([terms.html + ebookfiles.html](https://github.com/Gluejar/regluit/pull/1213/files#diff-5548b3e2f5983994d9773886713eec11d44503d0af430359da760bb2399bcc49)) — retirement paragraph, Campaign definition now naming all three types, legacy provisions retained (the two live campaigns still run under them).

- [ ] **3. Two dates** — proposals: "Date of last revision" → set to actual deploy date at merge; "As of July 2026" → keep (decision date; operationally true).

## Deploy notes

Migration `core.0030` — state-only, no DDL. #1217/#1218 (comments removal) chain on it structurally (`0031`/`0032`), so merge order is enforced by Django.

<details>
<summary>Full build & review record (commit-by-commit rationale, Codex findings, suite evidence) — collapsed 2026-08-09 for concision</summary>

[...]

Issue #1130, opened 2026-04-23
Remove Django comments functionality — spam magnet, Eric approved removal


rdhyee:
## Background

Per Eric's decision in the 2026-04-23 meeting with Raymond: **Django's built-in comments functionality is no longer valuable** — it's become a spam magnet generating low-signal traffic and occasional related maintenance burden. Remove it.

This is also related to the current Django 4.2 upgrade ([#1123](https://github.com/Gluejar/regluit/pull/1123)): `django_comments` has been a source of on_delete state-drift issues (see companion PR [#1126](https://github.com/Gluejar/regluit/pull/1126)). Removing it entirely is a cleaner simplification than continuing to carry the overlay-migration fixes.

## Eric's guidance

- OK to leave comments functionality **temporarily broken** while removal is in progress — better to accept a broken endpoint than to trigger more error-email floods via partial removal.
- Remove, don't gate behind feature flag — we don't want it back.

## Scope

- Remove `django_contrib_comments` (or whichever comments package is in use) from `INSTALLED_APPS`
- Remove `Comment`-related views, URLs, templates
- Remove `django_comments` migrations from the state-drift overlay ([#1126](https://github.com/Gluejar/regluit/pull/1126)) — no longer needed if the app is gone
- Data migration: decide whether to preserve historical comments (archive table) or drop them entirely (likely drop; low value, high spam content)
- Update templates that reference comment counts / comment forms

## Coordination with other PRs

- **After #1123 (Django 4.2) merges**: removal is easier in the 4.2 codebase than retrofitting into 1.11
- **Interacts with #1126**: if comments are being removed, #1126's `django_comments` on_delete fix becomes irrelevant. Either merge #1126 first (unblocks #1123) then remove comments cleanly, OR fold the removal into #1123 itself. Coordinate with #1123's merge plan.

## Related

- [regluit#1123](https://github.com/Gluejar/regluit/pull/1123) — Django 4.2 upgrade
- [regluit#1126](https://github.com/Gluejar/regluit/pull/1126) — django_comments on_delete overlay migration fix (may become obsolete)
- 2026-04-23 meeting with Eric

## Out of scope

- User-visible messaging about comments going away — probably not needed (comments section has been unused / spam-heavy)
- Replacement commenting system — not wanted

rdhyee:
Re-confirming sequencing per 2026-04-28 review:

- **#1126 stays in flight** as the prereq for #1123 (it's only 8 files of overlay migrations — trivial cost).
[...]

PR #1218, opened 2026-08-06
Remove Django comments (#1130) — phase 2 FULL cleanup: tables + notification history (alternative to #1220)

rdhyee:
Phase 2 of [#1130](https://github.com/Gluejar/regluit/issues/1130) — the **full-cleanup** option. Merges only after #1217 is deployed **and every process restarted**.

> **⚠️ Choose one: this PR or [#1220](https://github.com/Gluejar/regluit/pull/1220).** Both define a `core` `0032`. Git won't flag a conflict (different filenames) but Django will see two leaf migrations and refuse to migrate. Whichever is chosen, close the other.

## Which one you want depends on one question

**Is the goal to remove the comments *feature*, or to stop carrying comment *content*?**

The comments table is not where most of the comment content lives. Measured on production 2026-08-17:

| | |
|---|---|
| Comment text in `django_comments` | **127,848 characters** (468 rows, avg 273) |
| Comment text inside `Notice.message` | **23,810,322 characters** (22,120 rows, avg 1,076) |

**About 186× more comment-derived text sits in the notification history than in the comments table.** Every notification ever sent about a comment embedded a rendered copy of it.

- **#1220** drops the two tables and nothing else. Feature gone, history kept.
- **This PR** removes the comment-derived state broadly. Chosen under a simplification goal: the live database should stop carrying weight the site no longer uses.

## What this PR destroys (production figures, 2026-08-17)

| | Count |
|---|---|
| `django_comments` rows | 468 (130 authors, 0 anonymous) |
| `django_comment_flags` rows | 2 |
| `Notice` rows (cascade) | **22,120 of 45,291 — 49%**, spanning 2012-03-28 → 2025-03-13 |
| `NoticeSetting` rows (cascade) | 3,269 of 16,842 — 19% |
| `django_admin_log` rows (explicit) | 232 |
| `auth_permission` rows (cascade) | 16 |
| `NoticeType` rows | 3 |
| `ContentType` rows | 4 (`django_comments` ×2, ancient `comments` ×2) |
| `ObservedItem` rows | 0 — table is empty |

**This is irreversible.** Take a pre-deploy RDS snapshot and record its identifier. A snapshot restore is a last-resort preservation mechanism, not an ordinary rollback: separate instance, cutover required, post-snapshot writes lost. Exporting `django_comments` to a file first is cheap insurance at 468 rows.

## Operation order (revised after review)

1. **Scrub queued comment batches** — must precede the notice-type deletion, or `send_all()` raises on the missing type and wedges the queue
2. **Drop `django_comment_flags`, then `django_comments`** — the primary goal, FK order
3. **Delete the 232 admin-log rows** — explicitly, and *before* the ContentTypes
4. **Delete the 4 ContentType rows** — their 16 permissions cascade
5. **Delete the 3 NoticeTypes** — the largest cascade, deliberately last

The notice-type deletion moved to last so a partial failure can't leave the worst state — "notification history deleted, comment tables still present."

## Review fixes applied

- **Admin-log rows do NOT cascade.** `LogEntry.content_type` is `on_delete=SET_NULL` (verified on the deployed Django 4.2; `Permission.content_type` is `CASCADE`). Deleting the ContentTypes would have orphaned these 232 rows with `content_type_id = NULL` while leaving `object_repr` — a truncated copy of the comment text — in place. An earlier version of this migration claimed they cascaded. They're now deleted explicitly.
- **Hardened the batch scrub**, matching `core.0031` and the management command. It previously skipped non-conforming entries and returned `False`, suppressing the byte-scan fallback and potentially leaving a poisoned batch — the failure direction that wedges the whole queue. Non-conforming shapes now return `None`.
- **Removed every `noop` reverse.** Django previously reported this irreversible migration as reversible and would un-record it while the data stayed gone; it now raises `IrreversibleError`. `atomic = False` added, since MySQL can't roll back the DDL anyway.

## A correction to the record

The rationale on file was *"a spam magnet with low-value content."* **The production data doesn't support the spam half:** 0 anonymous comments (nothing got in unauthenticated), only 4 of 468 ever flagged or removed, 130 distinct authenticated authors, concentrated 2012–2014.

The *simplification* rationale stands on its own and is the basis for this PR. The spam one shouldn't be repeated.

## ⛔ Prerequisites before this runs

1. #1213 merged (phase 1's `core.0031` depends on its `0030_retire_pledge_b2u`)
2. #1217 deployed, **every Python process restarted**, post-restart scrub run, queue drain verified — repeated after any rollback
3. Pre-deploy RDS snapshot taken and its identifier recorded
4. #1220 closed

### DIFF
--- a/core/migrations/0032_remove_django_comments.py
+++ b/core/migrations/0032_remove_django_comments.py
@@ -35,7 +35,8 @@
 #    cascades to NoticeSetting / Notice / ObservedItem rows (RESTRICT FKs at
 #    the MySQL level; the ORM collector deletes children first).
 #
-# 3. Deletes the stale django_comments ContentType rows (cascades to their
+# 3. Deletes the stale ContentType rows for app labels django_comments and
+#    (ancient pre-1.6) comments (cascades to their
 #    auth_permission rows) so no dead metadata lingers in the admin.
 #
 # 4. Drops django_comment_flags then django_comments (FK ordering).
@@ -133,7 +134,12 @@ def delete_comment_notice_types(apps, schema_editor):
 
 def delete_comment_content_types(apps, schema_editor):
     ContentType = apps.get_model("contenttypes", "ContentType")
-    for content_type in ContentType.objects.filter(app_label="django_comments"):
+    # 'django_comments' is the modern app label; 'comments' rows are ancient
+    # django.contrib.comments (pre-Django-1.6) leftovers observed on the
+    # rehearsal box, still carrying add/change/delete permissions.
+    for content_type in ContentType.objects.filter(
+        app_label__in=["django_comments", "comments"]
+    ):
         content_type.delete()
 
 
--- a/core/migrations/0032_remove_django_comments.py
+++ b/core/migrations/0032_remove_django_comments.py
@@ -100,17 +100,34 @@ def _queued_comment_labels(pickled_data):
 
     Returns True/False from precise inspection of the label slot, or None
     if the pickle cannot be inspected.
+
+    "Cannot be inspected" MUST include any batch whose shape is not exactly what
+    queue() writes, not just batches that fail to unpickle. Returning False for a
+    non-conforming batch would claim "no comment notices here" and suppress the
+    caller's byte-scan fallback, leaving a poisoned batch in place. A byte-scan
+    false positive discards one pending batch; a false negative wedges the whole
+    notification queue indefinitely. (Matches the hardened copies in core.0031
+    and the scrub_comment_notice_batches management command.)
     """
     try:
         notices = _StubbedUnpickler(io.BytesIO(pickled_data)).load()
-        labels = set()
-        for entry in notices:
-            # queue() writes (user_pk, label, extra_context, on_site, sender)
-            if isinstance(entry, (tuple, list)) and len(entry) >= 2:
-                labels.add(entry[1])
-        return bool(labels & set(COMMENT_NOTICE_TYPES))
     except Exception:
         return None
+
+    # queue() pickles a list of (user_pk, label, extra_context, on_site, sender)
+    if not isinstance(notices, list):
+        return None
+
+    for entry in notices:
+        if not isinstance(entry, tuple) or len(entry) != 5:
+            return None
+        label = entry[1]
+        if not isinstance(label, str):
+            return None
+        if label in COMMENT_NOTICE_TYPES:
+            return True
+
+    return False
 
 
 def scrub_queued_comment_batches(apps, schema_editor):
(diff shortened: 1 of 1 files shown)


rdhyee:
**[CC]** Rehearsed on test.unglue.it 2026-08-06 in the intended order (phase-1 deployed + restarted first, then this migration): `core.0032` applied cleanly — comment tables dropped, 3 NoticeTypes deleted (38 others intact), ContentTypes deleted for both `django_comments` and the ancient pre-1.6 `comments` app label (discovered during rehearsal, migration extended accordingly), all comment permissions gone, notification queue at 0 and emitting normally. Full evidence on #1217. Remains **draft** until #1217 is merged and deployed.

🤖 *Posted by Claude Code on Raymond's behalf*

rdhyee:
**[Codex via CC]** Codex (gpt-5.4) review — **LGTM** (2026-08-17), under the simplification frame.

> **LGTM — safe to take out of draft as the chosen full-cleanup option.**

Three review rounds today. Codex initially returned **NOT-LGTM** on this PR, its lead objection being that deleting the notice types "unnecessarily destroys 22,120 historical Notice rows." That assessment assumed the goal was *remove the app, destroy no more than required*. Re-reviewed under the actual goal — **stop carrying state the site no longer uses** — it reversed:

> Yes, the NoticeType/Notice deletion is defensible under [the] stated goal. I no longer consider it wrong on the merits. If the objective is to remove unused comment-derived current state, retaining 23.8 million characters of rendered notification history — while deleting only 127,848 characters from the comments table — would largely miss that objective. Retaining the types also causes pointless NoticeSettings to keep being created. **This is deliberate data retirement, not accidental collateral damage.**

**Findings fixed across the rounds:**

1. **Admin-log rows do not cascade.** `LogEntry.content_type` is `on_delete=SET_NULL`, not `CASCADE` (verified on the deployed Django 4.2). The previous version would have orphaned 232 rows with `content_type_id = NULL` while leaving `object_repr` — a truncated copy of the comment text — in place. Now deleted explicitly, before the ContentTypes.
2. **Scrub false-negative.** `_queued_comment_labels()` returned `False` for non-conforming shapes, suppressing the byte-scan fallback and potentially leaving a poisoned batch — the direction that wedges the queue. Fixed here and in both #1217 copies. Tested across pickle protocols 0–5; the label-in-prose false-positive guard still holds.
3. **Misleading reversibility.** Every operation declared a `noop` reverse, so Django treated an irreversible migration as reversible. All removed; it now raises `IrreversibleError`. `atomic = False` added for MySQL DDL.
4. **Operation order.** The notice-type deletion moved last, so a partial failure can't produce "notification history deleted, comment tables still present." Codex confirmed the new order is safer and introduces no dependency hazard.
5. **Disclosure.** Header and PR body rewritten with the measured blast radius and the scope decision.

**Also corrected for the record:** the rationale on file was *"spam magnet, low-value content."* Production data does not support the spam half — 0 anonymous comments, 4 of 468 ever flagged or removed, 130 distinct authenticated authors, concentrated 2012–2014. The simplification rationale stands on its own.

**@eshellman — this is a genuine either/or:** this PR (full cleanup) or **#1220** (tables only, notification history kept). Both are reviewed and ready; they cannot both merge. The question is whether you want the comments *feature* gone or the comment *content* gone.

rdhyee:
**[CC]** Closing — **Eric chose the narrower option** at the 2026-08-17 meeting.

The decision, in his words as recorded: *"remove comment functionality, **leaving the notification history in the database**."*

So **[#1220](https://github.com/Gluejar/regluit/pull/1220)** is the phase-2 that ships: it drops the two comment tables (468 comments, 2 flag rows) and nothing else. The 22,120 notification records, 3,269 user notification settings, 232 admin-log rows, three notice types and four content types all stay.

Both PRs defined a `core` migration `0032`, so only one could ever merge. This one closes.

**Nothing here was wasted.** The review that produced this PR is what surfaced the real questions:

- Deleting the notice types would have cascaded to **22,120 notification records — 49% of all notifications, spanning 2012-03-28 to 2025-03-13**. That was documented in the migration but never quantified until this review, so the tradeoff had never actually been put to anyone.
- The **`SET_NULL` finding**: `LogEntry.content_type` doesn't cascade, so 232 admin-log rows would have survived with comment text still in `object_repr`. Confirmed in the wild — test.unglue.it, which ran an earlier version of this migration on 2026-08-06, has exactly 232 orphaned rows today.
- The **queue-scrubber false negative** — non-conforming batch shapes returned "no comments here" instead of falling through to the byte-scan. That fix was ported to both remaining copies in #1217, tested across pickle protocols 0–5, and re-tested against the old code, which missed 5 of 6 malformed-but-poisoned batches.
- The **`noop` reverses** that made Django treat an irreversible migration as reversible — also corrected in #1220.

Also worth leaving on the record: the rationale for removing comments was *"a spam magnet with low-value content."* Production data doesn't support the spam half — **0 anonymous comments**, only **4 of 468** ever flagged or removed, **130** distinct authenticated authors, concentrated 2012–2014. The comments still go; the reason is simplification, not spam.

Branch `comments-drop-migration-1130` retained for now in case the decision is ever revisited.

PR #1220, opened 2026-08-17
Remove Django comments (#1130) — phase 2 (minimal): drop comment tables only, replaces #1218

rdhyee:
**Replaces #1218.** Phase 2 of [#1130](https://github.com/Gluejar/regluit/issues/1130) — drops the comment tables and *only* the comment tables.

## Why a new PR

#1218 also deleted the three comment `NoticeType` rows and the comment `ContentType` rows. Measured against the **production database on 2026-08-17**, that cascade destroyed far more than the data actually being retired:

| Destroyed by #1218 | Count | Needed to remove the app? |
|---|---|---|
| `django_comments` rows | **468** (130 distinct authors, 0 anonymous) | ✅ yes — the point |
| `Notice` rows (cascade) | **22,120 of 45,291 — 49%**, spanning **2012-03-28 → 2025-03-13** | ❌ no |
| `NoticeSetting` rows (cascade) | **3,269 of 16,842 — 19%** | ❌ no |
| `django_admin_log` rows (cascade) | 232 | ❌ no |
| `auth_permission` rows (cascade) | 16 | ❌ no |
| `ObservedItem` rows | 0 (table is empty) | — |

**Keeping the three NoticeTypes keeps no dependency on the removed app.** A `Notice` stores its already-rendered `message` text plus a FK to its `NoticeType`; it holds no reference to a `Comment`. Nothing imports `django_comments` on their account, and no queued batch can be poisoned by their continued existence. Phase 1 ([#1217](https://github.com/Gluejar/regluit/pull/1217)) already stops those notices being created and filters the unreachable types out of the notification-settings UI.

Stale `ContentType` rows are inert. If they are ever worth removing, that is `manage.py remove_stale_contenttypes` — a deliberate, separately reviewable act, not a side effect of dropping two tables.

So: **this migration destroys 468 comments and their 2 flag rows, and nothing else.** Everything it does *not* do remains doable later; what it does do does not. That asymmetry is the whole argument for the smaller scope.

## Also fixes two review findings against #1218

- **Declares its irreversibility honestly.** No `reverse_sql`, so `migrate core 0031` raises `IrreversibleError`. #1218 used `RunSQL.noop`, which makes Django believe the migration is reversible and lets it un-record 0032 while the tables stay dropped. *(Verified on the deployed Django: `reversible` is `False` without `reverse_sql`, `True` with `noop`.)*
- **Drops the third copy of the pickle-scrub logic** rather than carrying and fixing it. Its role in #1218 was protecting queued batches from the `NoticeType` deletion; with that deletion gone, the hazard does not exist. The pickled-`Comment` hazard is entirely phase 1's, and is closed before this can run (migration `0031` clears the backlog; `manage.py scrub_comment_notice_batches` closes the migrate→restart window; after that restart no new comment can be posted).

  *(The false-negative path Codex identified in that scrub logic still needs fixing in the two remaining copies on #1217 — tracked separately, not a blocker for this PR.)*

## Schema facts (verified on production via `information_schema`)

- The **only** inbound FK to either table is `django_comment_flags.comment_id → django_comments`. Nothing else in the schema references them, so dropping flags then comments is necessary and sufficient.
- Outbound FKs (`auth_user`, `django_site`, `django_content_type`) are unaffected by dropping the referring tables.
- Both tables are InnoDB. `atomic = False` is explicit because MySQL cannot roll back DDL; if the play dies between the two drops, re-running converges (both use `IF EXISTS`).

## ⛔ Merge / deploy gate (unchanged)

Ships as **its own release**, only after #1217 is deployed **and every Python process restarted** (apache2, celeryd, celerybeat). That restart is what ends comment ingress. Shipping #1217 and this together would run both migrations with no restart between them and defeat the two-phase design.

**Take a pre-deploy RDS snapshot and record its identifier.** A snapshot restore is a last-resort preservation mechanism, not an ordinary rollback — separate instance, cutover required, post-snapshot writes lost. For selective recovery, export the two tables to a file first; it is 468 rows.

## Open question for @eshellman

Removing comments was agreed on the basis that they are *"a spam magnet with low-value content."* The production data doesn't show that: **0 anonymous comments** (nothing got in unauthenticated), only **4 of 468** ever flagged or removed, and the content is **468 comments from 130 real accounts**, concentrated in 2012–2014.

This PR is written to respect the decision while minimising what it costs. If you'd also like the 468 comments preserved before they go, an export step before deploy is trivial — say the word.

### DIFF
--- a/core/migrations/0032_drop_comment_tables.py
+++ b/core/migrations/0032_drop_comment_tables.py
@@ -1,0 +1,121 @@
+# Remove django-contrib-comments (#1130) — PHASE 2: drop the comment tables.
+#
+# ⛔ MERGE/DEPLOY GATE (unchanged from the previous phase-2 design):
+# this migration ships as its OWN RELEASE, after the comment-free code release
+# (PR #1217) has been deployed AND every Python process (apache2, celeryd,
+# celerybeat) has been restarted on the target box. That restart is what ends
+# comment ingress; until it happens, comment-aware code is still live and the
+# tables are still reachable.
+#
+# ─────────────────────────────────────────────────────────────────────────────
+# WHY THIS IS SMALLER THAN THE ORIGINAL PHASE 2
+#
+# The first phase-2 draft (PR #1218) also deleted the three comment NoticeType
+# rows and the comment ContentType rows. Measured against the production
+# database on 2026-08-17, that cascade was far larger than the data actually
+# being retired:
+#
+#     django_comments rows .................    468   (130 distinct authors)
+#     Notice rows destroyed by the cascade .. 22,120   of 45,291  (49%)
+#                                                      spanning 2012-03-28 → 2025-03-13
+#     NoticeSetting rows destroyed ..........  3,269   of 16,842  (19%)
+#     django_admin_log rows destroyed .......    232
+#     auth_permission rows destroyed ........     16
+#
+# None of that is required in order to remove django-contrib-comments:
+#
+#   * A `Notice` stores its ALREADY-RENDERED message text plus a FK to its
+#     NoticeType. It holds no reference to a Comment (see notification.models
+#     .Notice). Retaining the three NoticeTypes therefore keeps no dependency on
+#     the removed app — nothing imports django_comments, and no queued batch can
+#     be poisoned by their continued existence.
+#   * Phase 1 already stops those notices being created (the signal receiver and
+#     the comment_was_posted wiring are gone) and filters the now-unreachable
+#     types out of the notification-settings UI.
+#   * Stale ContentType rows are inert. If they are ever worth cleaning up, that
+#     is `manage.py remove_stale_contenttypes` — a deliberate, separately
+#     reviewable act, not a side effect of dropping two tables.
+#
+# So this migration destroys the 468 comments (and their 2 flag rows) and
+# nothing else. Everything it does NOT do is recoverable later; what it does do
+# is not. That asymmetry is the whole reason for the smaller scope.
+#
+# ─────────────────────────────────────────────────────────────────────────────
+# WHY THERE IS NO QUEUE SCRUB HERE
+#
+# The previous draft re-ran the phase-1 batch scrub as a "final belt". It was
+# there because deleting the NoticeType rows would break any queued batch that
+# still referenced those labels. Since this migration no longer deletes them,
+# that hazard does not exist.
+#
+# The remaining hazard — a queued batch containing a pickled
+# django_comments.Comment that can no longer be unpickled — is entirely phase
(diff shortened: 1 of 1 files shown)


rdhyee:
**[Codex via CC]** Codex (gpt-5.4) review — **LGTM** (2026-08-17)

> **Verdict: LGTM.** Safe to take out of draft. I found no code-level change required in this replacement. The migration correctly removes the two remaining runtime tables without repeating #1218's unnecessary destruction. Deployment still has hard operational prerequisites.

Codex separately reviewed **#1218** earlier today and returned **NOT-LGTM** on three findings — the 22,120-row `Notice` cascade, a false-negative path in the queue scrubber, and `noop` reverses that make Django treat an irreversible migration as reversible. This PR is the response to the first and third; the second is moot here because the scrub logic is not carried.

**Confirmed by review:**
- Phase 2's goal is met — retained ContentTypes, permissions, migration records and notification data are metadata or history, not dependencies
- Omitting the scrub is safe: dropping the tables doesn't change queue unpickling (the `Comment` class became unusable in phase 1), and the management command still works afterwards if a batch ever slips through
- Keeping the ContentTypes is safe — `model_class()` was already unavailable once the app left `INSTALLED_APPS`; admin log paths use stored representations and tolerate the absent model
- Drop order, `IF EXISTS`, `atomic = False`, the `0031` dependency, and `reversible == False` all independently verified

**Two review corrections applied in `915d8302`:**
- `remove_stale_contenttypes` **skips** apps removed from `INSTALLED_APPS`; cleanup needs `--include-stale-apps`
- The claim that no new poisoned batch can appear was too absolute — qualified with the three operational exceptions (phase-1 rollback, another old-code host sharing the database, a manual `queue()` call)

**Scope caveat worth stating plainly:** this is **not** complete erasure of comment content. Historical `Notice.message` values contain already-rendered copies of comment text and remain visible on the notices page. That is correct for *removing the application*, but would not satisfy a requirement to purge every surviving copy of the comments themselves. If the latter is the intent, say so and it needs a different (larger) change.

**Deployment prerequisites — not blockers to review, but hard gates before this runs:**
1. #1213's `0030_retire_pledge_b2u` must land first (phase 1's `0031` depends on it)
2. #1217 deployed **and every Python process restarted**, then the post-restart scrub, then queue-drain verified — repeated after any rollback
3. **#1218 must not merge.** Both define a `core` `0032`; Git won't flag a conflict (different filenames) but Django will see two leaf migrations and refuse to migrate. Close/supersede #1218 — do not paper over it with a merge migration
4. Pre-deploy RDS snapshot taken and its identifier recorded; optionally export the 468 comments first for selective recovery

PR #1223, opened 2026-08-23
Scrub regression tests + stale phase-2 comment fixes (folds into #1217)

rdhyee:
Folds into [#1217](https://github.com/Gluejar/regluit/pull/1217) (phase 1 of [#1130](https://github.com/Gluejar/regluit/issues/1130)) — closes the two release-readiness gaps found in the 2026-08-23 Codex release-plan review, ahead of the planned Mon/Tue production deploys.

## Why

1. **The scrub logic had no committed tests.** It was verified against synthetic pickle fixtures during review (documented in #1217's record) but nothing in the repo re-runs those checks. The logic exists in two deliberately-duplicated copies (migration `core.0031` + the management command) with no guard against divergence. Notably, the hardening in `25703468` had **never executed anywhere** — test.unglue.it runs an older cut of this branch.
2. **Two comments still described the rejected phase-2 design.** `0031` promised "the phase-2 migration (0032) re-runs the scrub… as a final belt," and the command docstring cited "core.0031/core.0032" inspection logic. That was #1218, which Eric rejected 2026-08-17. The chosen phase 2 ([#1220](https://github.com/Gluejar/regluit/pull/1220)) carries **no scrub and deletes no notice types** — the post-restart command is the *last* line of defense. In a sequence whose safety boundary is human-enforced, comments promising a safety net that no longer exists are an operational hazard, not cosmetics.

## What's here

- **`core/test_scrub_comments.py`** — 15 tests, parametrized over **both** copies of `_queues_comment_labels`:
  - genuine comment batch (each of the 3 labels), incl. one embedding an **unimportable class** — the actual poisoned-batch scenario (fixture self-verifies that plain `pickle.loads` raises)
  - non-comment batch kept; mixed batch flagged
  - comment label in message *text* but not the label slot → `False` (the byte-scan false-positive case)
  - the `25703468` hardening: non-list / wrong arity / non-str label / non-tuple entry → `None` (never `False`), so the byte-scan fallback runs — including an end-to-end case where the fallback is what catches a non-conforming batch that *does* carry a label
  - corrupt + truncated pickles → `None`; protocol-0 pickle inspectable
  - command end-to-end against real `NoticeQueueBatch` rows, output counts asserted; empty-queue no-op (production's expected state — queue measured 0 on 2026-08-23)
- **Comment corrections** in `0031` and the management command describing what #1220 actually does and doesn't do.
- **Merge of current `master`** (post-#1213) — required anyway: `core.0031` depends on #1213's `core.0030`, so the pre-merge branch couldn't even build a migration graph (`NodeNotFoundError` — the structural order guarantee firing as designed). The `core/tests.py` two-sided overlap auto-merged cleanly.

## Verification (local, Django 4.2.21 / Python 3.12 / MySQL 8)

- 15/15 new tests pass
- full migration chain `0001→0031` (merged order, incl. `0030_retire_pledge_b2u`) applies cleanly on a fresh test DB
- `regluit.core.tests`: failure set **identical** on this branch and `master` (9 pre-existing offline/live-API failures; none introduced)


rdhyee:
**[Codex via CC]** Codex (gpt-5.4) — **LGTM** (2026-08-23, round 2).

R1 (not-LGTM): one blocking gap — protocol 0–5 matrix not committed (only protocol 0 was); poisoned fixture accepted any load failure and didn't model the production-shaped Django failure. Codex independently probed both implementations at all protocols and found **no code defect** — coverage-only findings.

R2 after `e3582a06`: *"**LGTM.** The blocking finding is cleared."* — the committed matrix now runs every verdict at protocols 0–HIGHEST against both copies; the vanished-module fixture asserts `ModuleNotFoundError`; the new fixture faithfully mirrors `Model.__reduce__` (`model_unpickle` + `(app_label, model)`) and verifies the `LookupError` for `django_comments`, unit-level and end-to-end. Codex independently exercised the production-shaped fixture across protocols 0–5 with the expected results.

16/16 pass locally (Django 4.2.21 / MySQL 8).

Issue #1230, opened 2026-08-26
Consolidate donation funds: retire the Monographs Fund, keep only the General Fund


rdhyee:
**[CC]** *Filed on Raymond's behalf, capturing the decision at the 2026-08-26 Eric/Raymond call.*

> **TL;DR:** The donation page offers two funds; per today's call, we'll offer just the General Fund. Template + form change only — no migration, past donations keep their labels. **One question for @eshellman before work starts:** are past Monographs-fund donations *restricted funds* in FEF's accounting? If yes, say so — we'll check the reporting surfaces first. If no, this is a quick win.

**Why:** At today's call we decided to simplify the donation offering to a single **FEF General Fund**. The Open Access Monographs Fund's mechanism — matching contributions to qualified ungluing campaigns — describes a program the site no longer operates (new B2U campaigns are closed; only Thanks-for-Ungluing remains). Offering a fund choice adds friction and explanation to the page we most need to convert on, for a distinction that no longer has an operational meaning.

Part of the donation-path punch list with #1142 (Stripe modernization) and #1229 (donation discoverability). Small, independent, and a natural precursor to both: whichever UI those produce should only have to present one fund.

**Current state (master @ 2026-08-26):**

- `/about/funds/` (`frontend/templates/about_funds.html`) presents two funds with a `reason` radio (`monographs` — checked by default — vs `general`), plus several paragraphs explaining the Monographs Fund's campaign-matching mechanics (including a pointer at the rights-holder tools flow).
- `frontend/forms/__init__.py` — `DonationForm.reason` (`ChoiceField`, `DONATION_CHOICES`).
- `frontend/views/__init__.py` — `NewDonationView` passes `reason` through as `paymentReason`.
- `payment/models.py` — `Transaction.reason` stores the fund id; the `FUNDS` dict maps it to a display name (used e.g. in receipts/acknowledgements). Default already falls back to `"general"`.

**Proposed change (minimal):**

1. `about_funds.html`: drop the fund-choice radio and the Monographs Fund sections; one clear General Fund pitch + the donate form.
2. `DonationForm`: remove the `reason` choice from the user-facing form; new donations record `reason="general"`.
3. **Keep** `Transaction.reason` and the `FUNDS` mapping — historical transactions with `reason="monographs"` must continue to render correctly. No migration, no data change.
4. Check for other surfaces that mention the Monographs Fund (docs/FAQ templates, campaign pages) and update wording.

[...]

PR #1233, opened 2026-08-27
Stripe modernization, phase 1: Stripe.js v2 -> v3 Elements (#1142)

rdhyee:
Phase 1 of #1142 (the "cheap" option from the scoping survey — client-side field swap only, backend untouched). Replaces the hand-built card-number/CVC/expiry `<input>` fields with Stripe's own hosted Card Element, on both entry points that had them: the donation/pledge card form (`cardform.html`) and the "update your card on file" form (`manage_account.html`). Both share `stripe_stuff.html`'s JS, which is the actual rewrite: Stripe.js v2 → v3, callback-based `Stripe.createToken()` → the v3 promise-based version.

**Why:** direct fix for the 2026-08-26 incident. The old expiry fields used a literal `value="MM"`/`"YYYY"` trick (real text painted grey via JS, not an HTML placeholder) with no `autocomplete` attributes — exactly the shape that fooled 1Password into skipping them, producing a tokenization failure invisible to the donor and unlogged server-side. Stripe's Card Element owns its own field state inside an iframe; that specific fake-value trick isn't possible there. (Can't promise this closes every possible autofill-mismatch class — Stripe's own docs note Elements has its own separate quirks — but the exact 8/26 mechanism goes away, and a tokenization failure now surfaces through the Element's `change` event and a `.catch()` on `createToken()` into the existing `.payment-errors` box, not silently.)

**Backend untouched, deliberately**: `payment/stripelib.py`'s `Pay`/`make_account` already treat the token as an opaque `tok_xxx` string (see the existing test asserting `token.id[:4] == "tok_"`) — Stripe.js v3's `createToken(cardElement, extraData)` produces the same token shape v2 did, the documented supported path for adding Elements to a legacy Charges-API integration without migrating off Charges itself (that's the bigger, separate PaymentIntents migration — phase 2, not this PR).

**3 rounds of Codex review, `CC+Codex+LGTM`**:
- R1 (NEEDS CHANGES): missing `.catch()` on the `createToken().then()` chain — if `createToken()` itself rejected rather than resolving with `{error:...}`, nothing ran: button stuck disabled, donor sees nothing. Exactly the silent-failure shape this PR exists to kill. Fixed.
- R1: my SCSS edit (`#card-element` styling) never shipped — this project compiles SCSS to CSS as a manual, out-of-band step (`pledge.css` is a committed build artifact from 2023; `libsass` isn't installed anywhere; `manage.py compilescss` doesn't even import on the boxes). Confirmed by curling the live staging CSS after a `collectstatic` deploy: zero occurrences of `card-element`. Fixed by compiling with Dart Sass and committing the result — diffed byte-for-byte against the old file to confirm only cosmetic minifier differences elsewhere (e.g. `#FFFFFF`→`#fff`).
- R2 (LGTM with nits): confirmed both fixes close the gaps; caught a stray `sourceMappingURL` comment pointing at a nonexistent path in my compiled CSS. Fixed (`--no-source-map`).

**Live-tested on staging** (test.unglue.it) via real browser automation, more than once, including after fixes:
- Stripe Card Element renders correctly with matching CSS (`getComputedStyle` confirmed `border:2px solid rgb(214,221,224)`, white background — verified via a hard-reload after discovering Chrome's own HTTP cache was masking the CSS fix in an earlier check).
- Accepted Stripe's test card (`4242 4242 4242 4242` / `12/34` / `123`) with correct auto-advance, zero JS console errors, tokenized successfully, submitted to the server correctly (`Transaction` created with expected `reason='general'` fields, per Tuesday's #1230 work).
- Found and diagnosed a **separate, unrelated** issue while testing: staging's Stripe test secret key is expired (`Expired API Key provided: sk_test_rG...V8Ii`), blocking full charge completion on staging for *any* branch — confirmed via a clean, no-traceback error page (pre-existing error handling working correctly) and via the fact this PR touches zero Python/Stripe-API-calling code. Not fixed here; flagged separately, needs a Stripe dashboard / 1Password decision.

Out of scope, deliberately (both dead code, verified before excluding):
- `payment/templates/stripe.html` (`StripeView`) — only routed when `settings.DEBUG=True`, never true on staging or prod; also still Stripe.js v1.
- `test/campaigntest.py` — Python-2-only Selenium harness (`from urlparse import ...`), can't even import under the Python 3.12 stack already running; not collected by any test runner.


rdhyee:
**[Codex via CC]** Round 2 (final) verdict: **LGTM with nits.** Both round-1 gaps confirmed closed — `.catch()` now routes rejected `createToken()` promises through the same recovery helper as normal Stripe errors; compiled `pledge.css` contains the intended `#card-element` rule and the obsolete `#card_Number` selector is gone. Token contract (`stripe_token=token.id`) matches Stripe's documented Elements-tokenization pattern. One non-blocking nit (stray `sourceMappingURL` comment in the compiled CSS) — fixed, `acc76205`.

🤖 Posted by Claude Code at Raymond's direction.

PR #1234, opened 2026-08-29
Fix GET /accounts/logout/ 405 under Django 5.2 (#1145 follow-up)

rdhyee:
## Bottom line

Django 5.0 made `django.contrib.auth.views.LogoutView` POST-only
(`http_method_names = ["post", "options"]`, confirmed against the installed
`django==5.2.15`). The site's three "Sign Out" / "log out" links were still
plain GET `<a>` tags, so `/accounts/logout/` has been 405ing for every real
user since the Django 5.2 deploy on **2026-08-26 ~6:50am ET** — this was
already flagged as a known follow-up in #1145 ("the Sign Out link should
eventually become a POST form") and never finished. RY caught the live 405
on 2026-08-29.

## Files changed

- `frontend/templates/base.html` — the nav's "Sign Out" link (rendered on
  every page for a logged-in user). `<form>` is valid `<li>` content, so
  this nests a POST form directly in place of the old `<a>`.
- `frontend/templates/gift_user_error.html`, `pledge_user_error.html` — the
  "...you need to log out, and then log in as X" links. These sit inside a
  `<p>`, where `<form>` is **not** valid content (browsers auto-close the
  paragraph, splitting the sentence). Fixed with a hidden
  `<form id="..." style="display:none">` placed just outside the `<p>`,
  referenced by an inline `<button form="...">` — `<button>` is plain
  phrasing content, valid inside `<p>`, and HTML5's `form=` attribute
  associates it with a form elsewhere in the document. The `?next=` target
  moves to a hidden `<input name="next">` (`LogoutView`'s
  `RedirectURLMixin` reads `next` from POST body first, GET second, so
  this is a strict equivalent).
- `static/scss/sitewide4.scss` — two small `<button>` CSS resets
  (`button.link-button`, `button.notbutton`) so the new buttons render
  identically to the `<a>` links they replace.
- `static/scss/sitewide4.css` — recompiled from source (Dart Sass,
  `--style=compressed --no-source-map`, same convention as the
  `pledge.css` precedent in #1142). The diff carries the usual
  compiler-version cosmetic churn (color keyword → hex, decimal leading
  zeros, attribute-selector quoting) alongside the real additions —
  verified selector-by-selector that nothing existing was dropped.
- `libraryauth/tests.py` — new `TestLogoutView` regression guard.

## Verification summary

- Read Django's actual installed source (not memory) for `LogoutView.http_method_names`
  and `RedirectURLMixin.get_redirect_url()`.
- Ran the real Django template parser against all three edited templates — zero syntax errors.
- Real Chrome render of the actual compiled CSS: new "Sign Out" button pixel-identical to an
  untouched nav link side by side; the inline "log out" button did **not** split its paragraph;
  clicking each control fired a real POST with the correct fields (including the dynamic `next`
  value).
- Full `libraryauth` + `frontend` suite (47 tests) against a throwaway Docker MySQL 8 container
  (torn down after). Diffed against an identical run on unmodified `origin/master` in the same
  environment: same 3 pre-existing failures + 1 pre-existing error on both (a test brittleness in
  `RhPageTests` and a DB-at-import-time test-discovery fragility — neither touches logout or
  anything this PR changes). Zero new failures.
- New `TestLogoutView` proven to actually catch the bug: reran assertion (d) — nav renders a POST
  form, not a GET link — against the pre-fix templates (`git stash` the fix, keep the test) and
  confirmed it fails there, passes with the fix.

## How to verify after deploy

1. `curl -sD - -o /dev/null https://unglue.it/accounts/logout/` while logged out should still
   show `405` (GET is correctly still rejected — that's Django's contract, not a bug).
2. Sign in on unglue.it, click "Sign Out" in the nav — should log out and redirect to `/`,
   not 405.
3. Trigger the gift/pledge "wrong user" error page while logged in as the wrong user (e.g. open a
   gift-book link while signed in as someone else) and click "log out" in the sentence — should
   log out and land back on that same page (`?next=` honored), not 405 or a broken paragraph.
4. `curl -si -X POST https://unglue.it/accounts/logout/` (no CSRF token) should get `403`, not
   `405` — confirms the endpoint is now actually reachable via POST and CSRF protection is intact.

https://claude.ai/code/session_015Tj9QJjTGs75ZgJ2tdpQtP


rdhyee:
## Codex review — round 1

**Verdict: not LGTM** (2 findings, both addressed in `1182250` — see below)

1. **CSS parity bug**: the nav "Sign Out" `<button>` didn't render identically to the `<a>` it
   replaced — `#user_menu li:hover` and the dropdown's `line-height` override only targeted `a`,
   so the button kept dark text on hover and rendered 36px tall instead of the sibling rows' 26.25px.
   `font: inherit` in the button's own rule also silently clobbered `font-weight: bold` from the
   `header-text` mixin.
2. **Test didn't actually prove POST**: the regression test only checked that the logout URL
   appeared as a form `action` rather than an `href` — a form with no `method` (defaulting to GET)
   would have passed while still reproducing the 405.

Full Codex output:

> Verdict: not quite LGTM. The core POST/logout approach is correct and secure, but I would address two concrete issues before merge.
>
> 1. The nav button does not render identically to the old link... Move `font: inherit` before the mixin or restore `font-weight: bold`, and include `button.notbutton` alongside `a` in the dropdown hover and line-height rules.
>
> 2. The regression test does not actually prove the form uses POST... At minimum, assert: `method="post"`, a submit button is associated with that form, a CSRF input is present.
>
> Security and HTML conclusions:
> - CSRF handling is correct: every form includes `{% csrf_token %}`, and Django 5.2's `LogoutView` is independently decorated with `csrf_protect`.
> - The hidden POST `next` does not introduce an open redirect. `RedirectURLMixin` reads POST before GET, validates the destination against the request host and HTTPS requirement, and otherwise falls back to `LOGOUT_REDIRECT_URL`.
> - The `form` cross-reference is spec-compliant: `button` is phrasing content permitted inside a paragraph, and its `form` attribute may identify a form elsewhere in the same document. It works in all current browsers.
> - IE11 and Edge 12–15 do not support this association. If legacy-browser support matters, the simpler alternative is `<form><p>…<button>…</button>…</p></form>`.
>
> One separate, pre-existing concern: `registration_base.html` stores an arbitrary raw `next` query value in a cookie and later feeds it to `window.location.replace()`. That potentially bypasses Django's server-side validation and deserves a security follow-up, but this PR neither introduces nor worsens it.

Both findings addressed in `1182250`: CSS specificity fixed (verified against the actual
compiled output), test strengthened with BeautifulSoup assertions for `method="post"` + CSRF +
submit control, plus new template-level tests for the two error pages and an explicit
open-redirect regression test. Full libraryauth + frontend suite (50 tests, throwaway Docker
MySQL) still shows only the same pre-existing failures as unmodified master.

Noted the pre-existing `registration_base.html` cookie/`window.location.replace()` concern as
out of scope for this PR (doesn't touch anything this fix changes) — worth a separate follow-up.

rdhyee:
## Codex review — round 2

**Verdict: LGTM — no remaining findings.**

> - CSS parity is correct. The compiled CSS gives the `#user_menu` hover and line-height selectors greater specificity than the base button rule, producing white hover text and `26.25px` line-height. The narrowed font reset preserves bold weight from `header-text`.
> - The nav regression test now requires `method="post"`, a nonempty CSRF token, and a submit control.
> - The error-template tests correctly verify external button/form association, POST action, CSRF presence, `next`, and absence of a wrapping form.
> - The safe and external `next` tests adequately cover redirect behavior.
> - `git diff --check` passes, and BeautifulSoup is already pinned in the project dependencies.
>
> I inspected the committed compiled CSS but couldn't independently rerun Django tests locally: this checkout's `.venv` lacks Django and the sandbox cannot access Docker. That doesn't change the code-review verdict, given the source/compiled-artifact checks and your reported MySQL suite comparison.

Both round-1 findings confirmed fixed. Codex is LGTM as of `1182250`. Moving to a final CC review pass next.

rdhyee:
## CC final review pass

Ran my own review (code-review skill, high effort) on top of Codex's round-2 LGTM. It surfaced
three candidate findings; I verified each against the actual running code rather than taking them
at face value:

1. **False positive** — claimed `button.notbutton` (Sign Out) lacks a `:hover` underline that its
   `<a class="notbutton">` siblings (My Faves, Notices, ...) have. Checked with
   `getComputedStyle()` in a real Chrome render: **none** of the nav items get a hover underline,
   fixed or unfixed — a pre-existing `.js-topmenu ul li a { text-decoration: none }` rule (higher
   specificity than the global `a:hover{underline}` rule) already suppresses it site-wide. Also
   reconfirmed color/font-weight/line-height are pixel-identical between the button and its `<a>`
   sibling on hover. No change needed.
2. **Judged non-blocking** — the hidden-form-plus-`button[form=]` pattern is duplicated across the
   two error templates (2 call sites). Per this repo's own convention (prefer a narrow fix over
   speculative abstraction), left as-is.
3. **Real, fixed** — `test_post_rejects_external_next_as_open_redirect` only asserted the redirect
   didn't *contain* the evil hostname, not that it landed on the actual safe fallback. A subtler
   bypass not literally containing that substring would have slipped through. Tightened to
   `assertRedirects(resp, '/', ...)` against the app's real `LOGOUT_REDIRECT_URL`.

## Codex review — round 3 (final)

Sent Codex the small follow-up diff since it changed code after their round-2 LGTM:

> Yes—LGTM overall at `3d3bcf9b`.
>
> The new `assertRedirects(resp, '/', fetch_redirect_response=False)` correctly verifies the exact safe fallback and subsumes the prior status-code check. The PR head is confirmed at that commit, mergeable and clean. No new concerns.

**Both CC and Codex are LGTM at `3d3bcf9b`.** Full libraryauth + frontend suite (50 tests, throwaway
Docker MySQL 8) unchanged throughout: same pre-existing 3 failures + 1 error as unmodified
`origin/master`, zero new failures at every round. Applying `CC+Codex+LGTM`.

PR #1235, opened 2026-08-29
Release: Fix GET /accounts/logout/ 405 under Django 5.2 (#1234 / #1145)

rdhyee:
Release: **[#1234](https://github.com/Gluejar/regluit/pull/1234)** — closes the #1145 follow-up.
Sole content of this release.

**Change:** Django 5.0 made `LogoutView` POST-only. The site's three "Sign Out" / "log out" links
were still plain GET `<a>` tags, so `/accounts/logout/` has 405'd for every real user since the
Django 5.2 deploy on 2026-08-26 (~3 days). Converted all three to POST forms with `{% csrf_token %}`:
`base.html`'s nav (`<form>` nested directly, valid `<li>` content), and the two error-page links
that sit inside a `<p>` (`<form>` isn't valid `<p>` content — used a hidden `<form>` outside the
paragraph plus an inline `<button form="...">`, the HTML5 form-association attribute). Two small
CSS resets in `sitewide4.scss` keep the new `<button>` elements pixel-identical to the `<a>` links
they replace.

**Static assets changed** (`static/scss/sitewide4.css`, recompiled from source) → deploy needs
`-e run_collectstatic=true`. **No migration, no dependency change.**

**Rollback:** redeploy `1b191e6d` (current `production` tip).

**Verification before this release:**
- 3 rounds of Codex review + a final independent CC pass → `CC+Codex+LGTM` on #1234. Two rounds
  found and fixed real issues (a CSS hover/font-weight parity bug, a regression test that didn't
  actually assert POST); the CC pass found and fixed one more (a weak open-redirect assertion) and
  verified two other candidate findings were false positives/non-issues in a real browser before
  dismissing them.
- Full `libraryauth` + `frontend` suite (50 tests, throwaway Docker MySQL 8) at every round: same
  pre-existing 3 failures + 1 error as unmodified `origin/master`, zero new failures.
- New `TestLogoutView` / `TestLogoutFormsInErrorTemplates` regression tests — proven to actually
  catch the original bug (reran against the pre-fix templates and confirmed they fail there).
- **Deployed and live-tested on test.unglue.it** (this session, `3d3bcf9b`): `GET /accounts/logout/`
  still correctly 405s; `POST` without a CSRF token correctly 403s (proving the endpoint is now
  reachable and protected, not just permanently blocked); signed in with a real account and
  clicked "Sign Out" in the nav — logged out cleanly, no 405, landed on the logged-out homepage.
  (One non-issue found along the way: the button briefly rendered with default browser chrome
  until a hard refresh — a stale browser cache of the CSS file, pre-existing behavior of how this
  app serves static assets, unrelated to this fix and confirmed gone after a real reload.)
- The two gift/pledge "wrong user" error-page links are covered by automated markup tests
  (`TestLogoutFormsInErrorTemplates`: correct `method="post"`, CSRF, `next=`, button↔form
  association) and share the exact mechanism just proven live for the nav, but were not
  live-clicked on staging (would need a real mismatched gift/transaction to trigger).

https://claude.ai/code/session_015Tj9QJjTGs75ZgJ2tdpQtP


rdhyee:
## Deployed to production — 2026-08-29 14:41 UTC

`deploy.yml -e run_collectstatic=true` → **`production @ a64d6f8b`** (matches this PR's merge commit). 32s, `ok=7 changed=5 failed=0`.

**Verified live on unglue.it immediately after:**
- `GET /accounts/logout/` → `405 Allow: POST, OPTIONS` (correct — Django's contract, unchanged)
- `POST /accounts/logout/` with no CSRF token → `403` (was `405` before this deploy; proves the endpoint is now reachable and CSRF-protected)
- `/static/scss/sitewide4.css` → fresh `Last-Modified: 14:41:38`, UTF-8 BOM present, all 3 `button.notbutton` rules + `button.link-button` present
- `/`, `/accounts/login/`, `/accounts/superlogin/`, `/faq/` → all `200`; Google sign-in still offered
- Apache `error.log`: **0 tracebacks** since the deploy (only the 3 restart lines), re-checked after the real logout below
- **Real browser test**: hard-reloaded unglue.it as a logged-in user, opened the nav menu ("Sign Out" rendered identically to its sibling rows — no border, bold, 26.25px), clicked it → logged out cleanly, no error, landed on the logged-out homepage

**Known cosmetic-only lag, pre-existing:** static files are served with no `Cache-Control`/`Expires`, so browsers guess freshness from `Last-Modified` — users who cached the old CSS recently may see an unstyled (but fully working) Sign Out button for up to ~7 days until their copy expires. Same as every prior CSS change on this site; a hard refresh clears it. Not introduced by this release.

**Rollback if ever needed:** `ansible-playbook -i hosts deploy.yml -e git_branch=1b191e6d -e run_collectstatic=true`

Issue #1238, opened 2026-08-31
Staging-refresh playbook: prod-DB-copy boxes must not leak real user emails or links


rdhyee:
## TL;DR

**The short version:** staging (`test.unglue.it`) sometimes gets a straight copy of production's database. That's the root cause of two problems: password-reset links pointing at the wrong site (fixed live, but keeps coming back), and — the bigger one — nothing stops staging from actually emailing real users if someone triggers a mail-sending action while poking around. RY's preferred fix: don't copy raw prod data onto staging in the first place, use a scrubbed/synthetic copy instead (bigger, separately-scoped work). In the meantime, [PR #1239](https://github.com/Gluejar/regluit/pull/1239) adds a backstop — an opt-in switch that redirects any staging email to a safe address instead of a real one — and is in final review now.

---

## Why this is filed now

Found live on `test.unglue.it` 2026-08-31 while verifying [#1237](https://github.com/Gluejar/regluit/pull/1237): a password-reset request emailed a link pointing at **prod** (`unglue.it`) instead of the box that actually sent it. Root cause: the box's database is a snapshot of production (per the internal ops note "Staging swap 2026-08-23" -- point-in-time prod copy), and the copy carried prod's `django.contrib.sites` row with it. This is a **recurrence** of #1164 (closed June, same bug on `dj42.unglue.it` after an earlier prod-snapshot restore) -- fixed live this time with the existing `manage.py set_site_domain` command, but it will keep coming back on every future snapshot restore because nothing runs that command automatically outside a full Ansible playbook pass, and manual/partial deploys (the common case on this box, see below) skip it.

Investigating that led to a bigger, **not-yet-tracked** finding: **test.unglue.it can send real transactional email to real users.** No `EMAIL_BACKEND` override exists anywhere for non-prod (`settings/common.py`, `settings/dev.py`, `regluit-provisioning/group_vars/test/vars.yml` -- checked all three). The box uses the real SES-backed SMTP relay (`smtp.gluejar.com`), and once its DB is a prod snapshot, that means **every real user's real email address is live in a system that will actually mail them** -- password resets, gift notifications, campaign/notice emails, anything a human or a script triggers while poking at staging. This isn't hypothetical: `regluit-provisioning#22` (closed) shows staging email already surprised Eric once, when a leftover celery test-task fired every 5 minutes; that fix removed one specific task, not the underlying gap.

## What's already tracked (don't duplicate -- this issue is the missing piece, not a replacement)

[...]

PR #1237, opened 2026-08-30
Remove Sign-in-with-Google (Eric: unused OAuth client expiring)

rdhyee:
## Summary

- Fixes #1236. Eric, verbatim, replying to Google's unused-OAuth-client notice: "lets remove the log in with google button." (Gmail thread `1a04cd3f47fac2a2`, 2026-08-30.)
- Removes the "Sign in/Sign Up with Google" link from every page that had one (`registration/login.html`, `home.html`, `gift_login.html`, `registration/from_pledge.html`, `registration/registration_form.html`) and deregisters `social_core.backends.google.GoogleOAuth2` from `AUTHENTICATION_BACKENDS`, plus the now-dead dummy Google OAuth2 key/secret settings.
- Deliberately scoped to Google only: `social_django`, its URL include, the generic social-auth pipeline, and the still-registered (if unused) `OpenIdAuth` backend are untouched. Existing Google-linked users' `UserSocialAuth` data and the account page's use of it are untouched -- those users keep their accounts and can still log in with a password.
- Adds `libraryauth/tests.py::TestGoogleLoginRemoved` as a regression guard.

## User-facing effect

A Google-linked user sees no functional change to their account -- they're still logged in via username/password like everyone else once the button is gone; they just lose the one-click Google option and would use "Forgot your password?" if they never set one. No accounts or data are deleted.

## Test plan

- [x] `manage.py test libraryauth frontend` locally (Python 3.12.9 venv + disposable Docker MySQL 8) -- 56 tests, 3 failures + 2 errors, diffed byte-for-byte identical (same tests, same failure reasons) against an unmodified `origin/master` baseline run side-by-side in a throwaway worktree. Zero net-new failures. All 6 new `TestGoogleLoginRemoved` tests pass.
- [x] Manually confirmed no remaining references to `google-oauth2` / `btn-google-plus` / `GoogleOAuth2` anywhere in the codebase (`grep`).
- [ ] Deploy to test.unglue.it and click through login/home/registration pages (tracked separately, RY-requested).

No prod deploy in this PR -- humans merge/deploy per usual.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_015uv8CgcRqX4qjx9C74bVHj

Issue #1164, opened 2026-06-16
Staging boxes emit prod (unglue.it) links in password-reset & other emails — Site row not localized after snapshot restore


rdhyee:
## Symptom

Requesting a password reset on **dj42.unglue.it** produces an email whose link points at **prod**:

> To reset your password, please click the following link …
> `https://unglue.it/accounts/reset/MQ/…/`

…even though the request was made on `dj42.unglue.it`. (Observed 2026-06-15, RaymondYee's account.)

## Root cause (confirmed on the box)

The templates are correct — `frontend/templates/registration/password_reset_email.{html,txt}` render:

```
{{ protocol }}://{{ domain }}{% url 'password_reset_confirm' uid token %}
```

`{{ domain }}` / `{{ site_name }}` are supplied by Django's `PasswordResetView` via `get_current_site(request)`. Because `django.contrib.sites` is installed (`settings/common.py` → `'django.contrib.sites'`, `SITE_ID = 1`), `get_current_site()` returns the **DB `Site` row for `SITE_ID`**, *deliberately ignoring the request host*.

dj42 was restored from a prod snapshot, so its `Site` row #1 still carries prod's domain. Verified via `django-admin shell` on dj42:

```
SITE_ID = 1
1 'unglue.it'           'unglue.it'
2 'please.unglueit.com' 'unglue.it development'
3 'localhost:8000'      'unglue.it local development'
4 'ry-dev.unglueit.com' 'ry-dev development'
5 'just.unglue.it'      'unglue.it staging'
current = unglue.it
```

So `current_site.domain == 'unglue.it'` on dj42 → the reset link points at prod.

## Scope — broader than password reset

This affects **every** Sites-derived outbound URL/label, not just reset. Same `Site.objects.get_current()` is used in:
- `core/signals.py` (notice emails)
- `payment/tasks.py` (payment/receipt emails)
- `frontend/views/__init__.py`, `core/models/__init__.py` (ask_pdf/ask_epub, gift, etc.)

A staging user who clicks any emailed link lands on prod.

## NOT a cutover blocker, NOT a security bug

- On **real prod**, `Site #1 = unglue.it` is *correct*. This is purely a staging-localization gap for boxes restored from prod snapshots.
- Pinning reset links to the Sites domain (rather than `request.get_host()`) is Django's **recommended defense against Host-header poisoning** of reset emails. The fix is to localize the staging Site row, *not* to switch to the request host.

## Recommended fix

[...]

PR #1239, opened 2026-08-31
Add opt-in email allowlist/redirect backend for non-prod (backstop for #1238)

rdhyee:
## TL;DR

**The problem:** when a staging server's database gets refreshed from a copy of production, it inherits every real user's real email address — and nothing stops it from actually mailing them (password resets, notices, whatever a human or script triggers while testing). **The fix:** an opt-in switch (off by default, zero effect on production) that redirects any staging email to a safe address instead, so testers can still see what would have gone out without a real person ever receiving it.

## Summary

- Part of #1238 (staging-refresh playbook). Found live on test.unglue.it 2026-08-31 while verifying #1237: nothing stops a staging/test box whose database has been refreshed from a prod snapshot from sending real transactional email (password resets, gift notices, campaign emails) to real users' real addresses -- confirmed by checking every settings module and `regluit-provisioning/group_vars/test`, none of which override `EMAIL_BACKEND`.
- Adds `utils/safe_email_backend.py`: `AllowlistEmailBackend` wraps the real backend and, per message, either passes it through unchanged (every recipient allowlisted) or redirects the whole thing to a configured catch-all, with the original recipients recorded in the body (subject stays generic -- see review history below) -- never silently drops (that's exactly what made #1164 and regluit-provisioning#22 harder to diagnose than necessary) and never silently delivers to an unknown address (refuses/raises if no redirect target is configured).
- Wired behind `EMAIL_SAFE_MODE` (env var, default off) in `settings/common.py` -- zero effect on production or any environment that hasn't explicitly opted in.

## Scope note

This is a backstop, not the primary fix. RY's stated preference (see #1238) is a synthetic/scrubbed test DB instead of a raw prod copy in the first place -- a bigger, separately-scoped follow-up. This backend is worth having regardless, for whatever a scrub script misses or a deliberate one-off raw-copy debugging session.

Deploying this to test.unglue.it (setting `EMAIL_SAFE_MODE=true` + an allowlist/redirect address there) is a separate, provisioning-side step, not done in this PR.

## Review history (worth reading if you're skimming past the label)

Three rounds of Codex review, each finding real bugs -- see the PR comment for the full breakdown. Highlights: a message subclass could have hidden a recipient from the allowlist check entirely (fixed by checking Django's own `recipients()` rather than `.to`/`.cc`/`.bcc`); the safety-net's own "refuse and log loudly" path was itself silently disabled by a Django logging-config gap; a malformed address value could fail open instead of closed; and the redirect notice went through two more passes before it stopped leaking recipient PII into logs/subjects. All fixed and re-verified live, not just re-tested.

## Test plan

- [x] `manage.py test libraryauth frontend utils` locally (Python 3.12.9 venv + disposable Docker MySQL 8) -- 74 tests, 3 failures + 2 errors, the same pre-existing baseline as origin/master (established while verifying #1237 in the same environment). Zero net-new failures across every commit in this PR.
- [x] 24 tests in `utils/tests.py` covering: allowlist matching (domain/address/case-insensitivity/empty-recipient/malformed-value/display-name), full pass-through, redirect on any non-allowlisted recipient (including mixed To lists, Cc, and Bcc), the `recipients()`-override bypass, context-manager support, refusal to send with no redirect target configured, the refusal log carrying no PII, and that `EMAIL_SAFE_MODE` defaults off.

https://claude.ai/code/session_015uv8CgcRqX4qjx9C74bVHj

### DIFF
--- a/settings/common.py
+++ b/settings/common.py
@@ -569,13 +569,25 @@ DOWNLOAD_LOGS_MAX = 499
 # (password resets, gift notices, campaign emails, ...) would deliver to
 # those real people. Off by default (empty string) so production, and any
 # environment that hasn't explicitly opted in, are unaffected.
+#
+# IMPORTANT if you're touching settings after this point: this must be the
+# LAST thing in the settings chain to set EMAIL_BACKEND. Any settings module
+# that does `from .common import *` and then re-sets EMAIL_BACKEND itself
+# (settings/spike.py already does exactly this, for an unrelated reason)
+# silently defeats this safety net. Verified 2026-08-31 that the actual
+# deploy template (regluit-provisioning/roles/regluit_prod/templates/
+# prod.py.j2, which is what test.unglue.it/unglue.it actually run as
+# regluit.settings.prod) does NOT re-set EMAIL_BACKEND after `from .common
+# import *` -- but that's an external file this repo doesn't control, so it
+# stays a live risk for any future change there. (CC review, 2026-08-31.)
+from regluit.utils.safe_email_backend import resolve_email_backend  # noqa: E402
+
 EMAIL_SAFE_MODE = os.environ.get('EMAIL_SAFE_MODE', '').strip().lower() in ('1', 'true', 'yes')
+SAFE_EMAIL_REAL_BACKEND, EMAIL_BACKEND = resolve_email_backend(
+    EMAIL_SAFE_MODE,
+    globals().get('EMAIL_BACKEND', 'django.core.mail.backends.smtp.EmailBackend'),
+)
 if EMAIL_SAFE_MODE:
-    SAFE_EMAIL_REAL_BACKEND = os.environ.get(
-        'SAFE_EMAIL_REAL_BACKEND',
-        globals().get('EMAIL_BACKEND', 'django.core.mail.backends.smtp.EmailBackend'),
-    )
-    EMAIL_BACKEND = 'regluit.utils.safe_email_backend.AllowlistEmailBackend'
     EMAIL_SAFE_MODE_ALLOWED_DOMAINS = os.environ.get('EMAIL_SAFE_MODE_ALLOWED_DOMAINS', '')
     EMAIL_SAFE_MODE_ALLOWED_ADDRESSES = os.environ.get('EMAIL_SAFE_MODE_ALLOWED_ADDRESSES', '')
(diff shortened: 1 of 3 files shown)


rdhyee:
## Review summary: CC + Codex, both LGTM

Three rounds of Codex review, each finding real bugs -- all fixed and re-verified live (not just re-tested):

**Round 1**: (a) HIGH -- checked `message.to/.cc/.bcc` directly instead of `message.recipients()`, letting a message subclass with an overridden `recipients()` add a hidden recipient that bypassed the allowlist entirely. (b) MEDIUM -- didn't inherit Django's `BaseEmailBackend`, so `with backend:` raised `TypeError`. Both fixed in `68dcdb75`.

**Round 2**: (a) MEDIUM -- the ERROR log meant to make a Celery-swallowed send-refusal visible was itself silently disabled (`disable_existing_loggers=True` + no explicit `LOGGING['loggers']` entry). Verified live: `logger.disabled` was `True` before the fix, `False` after. (b) MEDIUM -- `_is_allowed()` had a fail-open bypass: `parseaddr('real@x.com, ok@allowed.org')` returns `('', '')`, which was treated the same as "no recipient here." (c) LOW -- log/exception text embedded actual recipient addresses. All fixed in `f8ba2f94`.

**Round 3**: the round-2 fix for (c) replaced the address list with the message subject as a "safe" identifier -- but a subject can itself carry PII (e.g. "Password reset for real@example.com"). Dropped entirely; the log line is now a bare recipient count, verified via `assertLogs` capturing the actual emitted record. Fixed in `8538ea09`. One suggestion (full RFC 5322 mailbox validation vs. `parseaddr`'s looser parsing) was left as a documented, deliberate stopping point -- no concrete example surfaced where it lets a genuinely non-allowlisted address through, only allowed addresses in unusual forms.

**CC review**: the code-review skill's automated fork stalled for several hours (spawned an unusually large multi-angle sub-team and never converged) -- rather than keep blocking on it, did a direct line-by-line read of the final diff myself, focused on exactly the failure classes Codex had been finding (recipient-list handling, settings-loading order, logging config, PII placement). Found nothing further.

Verified: `manage.py test libraryauth frontend utils` -- 74 tests, 3 failures + 2 errors, the same pre-existing baseline as origin/master (established in #1237's session, re-confirmed here) -- zero net-new failures across every commit in this PR.

PR #1241, opened 2026-08-31
Guard login form against password-manager double-submit (403 CSRF on login) (#1240)

### LINKED ISSUE #1240: CSRF 403 on password login when a password manager auto-submits the form (double-submit race)
**[CC]**

## Summary

Password login on unglue.it returns a **403 CSRF failure page even though the login actually succeeded**, whenever a password manager auto-submits the login form and the user then clicks "Sign in with Password" themselves. The user sees a scary 403; behind the scenes they are logged in.

## Evidence (2026-08-31, prod)

Apache access logs show every successful password login today paired with a 403 one second later, all from the same client:

```
19:45:56 POST /accounts/superlogin/ 302   ← login succeeds, CSRF cookie rotates
19:45:57 POST /accounts/superlogin/ 403   ← duplicate, stale token
19:46:16 POST /accounts/superlogin/ 302
19:46:16 POST /accounts/superlogin/ 302   ← both raced in same second: both valid
19:52:06 POST /accounts/superlogin/ 302
19:52:07 POST /accounts/superlogin/ 403
...
```

Live reproduction with an instrumented page (capture-phase `submit`/`click` listeners logging to localStorage) caught the mechanism directly:

| time | event | `isTrusted` |
|---|---|---|
| t+0.000s | click on "Sign in with Password" | **false** — synthetic (password-manager auto-submit) |
| t+0.000s | form submit → POST #1 → 302, login succeeds, Django rotates CSRF cookie |  |
| t+1.178s | click on "Sign in with Password" | **true** — the human |
| t+1.178s | form submit → POST #2 with pre-rotation token → **403** |  |

The captured cookie/token pair from the 403 genuinely fails `django.middleware.csrf._unmask_cipher_token()` — it is a real mismatch, caused by the rotation from POST #1, not a Django bug.

## Why this surfaced only now

Prod logs show **zero** successful `superlogin` POSTs (302s) in the preceding two weeks — password login is almost never exercised (users, including admins, used Google sign-in). With Sign-in-with-Google being removed (#1237), password login is about to become the *only* login path, so this failure class will hit real users.

## Proposed fix

1. **Double-submit guard on the login form** (this issue's PR): first submission marks the form; any second submission is prevented. The password manager's submit wins; the human's follow-up click becomes a no-op and the browser completes the successful login navigation. Small, template-local, no server change.
2. **Follow-up (optional)**: a friendlier `CSRF_FAILURE_VIEW` — if a login POST fails CSRF but the user is already authenticated, redirect onward instead of rendering a bare 403.

User-side note: disabling the password manager's "auto-submit after fill" also avoids the race, but we can't ask that of every visitor.

---
🤖 Investigated and filed by Claude Code with Raymond.
https://claude.ai/code/session_01EqzfARuryr6m6jzGscgAY6

rdhyee:
**[CC]**

Fixes #1240.

## Why

Password login shows a bare **403 CSRF page even though login succeeded**, whenever a password manager auto-submits the form and the user also clicks the button. Root cause (confirmed live on prod, 2026-08-31, via instrumented capture-phase listeners): the manager fires a synthetic click (`isTrusted: false`) ~1.2s before the human's real click. POST #1 logs in and Django rotates the CSRF cookie; POST #2 carries the stale token → genuine mismatch → 403. With Sign-in-with-Google going away (#1237), password login becomes the only path, so this will hit real users.

## What

- `static/js/sitewide1.js`: document-level delegated `submit` listener scoped to forms posting to `/accounts/superlogin/`. First submission marks the form (`data-login-submitted`); any further submission while it's in flight is `preventDefault()`ed, so the manager's successful login navigation completes. The button is disabled a tick later (after form data capture) as visual feedback, and the guard self-resets after 8s in case navigation never happens (network failure) so the form can't get stuck.
- `frontend/tests.py`: two smoke tests pinning the wiring — the login page loads `sitewide1.js`, the form posts to the guarded action, and the script contains the guard.

## Why sitewide JS and not the form template

The sign-in lightbox is injected via `$j("#lightbox").load(url + " #lightbox_content")` — jQuery **strips inline `<script>` tags when loading with a fragment selector**, so a script inside `login_form.html` would never run in the modal, which is exactly where the bug was hit.

## Known limits (deliberate)

- A manager that calls `form.submit()` directly from its isolated world bypasses `submit` events and this guard. The observed behavior (1Password-style synthetic click) does fire the event and is covered.
- The registration form in `from_pledge.html` shares `class="login"` but a different action; the guard is scoped by action, so it is untouched.
- Follow-up idea in #1240: friendlier `CSRF_FAILURE_VIEW` when the user is already authenticated.

## Test plan

- [x] `manage.py test regluit.libraryauth regluit.frontend regluit.utils` (py3.12 venv + disposable Docker MySQL 8): 76 tests = 74 baseline + 2 new; same 3 pre-existing failures + 2 pre-existing errors as origin/master baseline (RhPage csrf ×3, registration, views loader — all environmental). Zero net-new failures.
- [x] `node --check` on the modified JS.
- [ ] Live verification on test.unglue.it by RY (password-manager autofill + manual click; **hard-reload needed** — static files are unhashed, so a cached `sitewide1.js` will mask the fix).

https://claude.ai/code/session_01EqzfARuryr6m6jzGscgAY6

### DIFF
--- a/frontend/tests.py
+++ b/frontend/tests.py
@@ -444,11 +444,13 @@ class CampaignRetirementTests(TestCase):
 
 
 class LoginDoubleSubmitGuardTests(TestCase):
-    """Smoke tests pinning the #1240 double-submit guard in place.
+    """Tests pinning the #1240 double-submit guard in place.
 
-    The guard itself is client-side JS (static/js/sitewide1.js); these tests
-    assert the wiring that makes it effective: the login form posts to the
-    URL the guard watches, and every page loads the script that carries it.
+    The guard itself is client-side JS (static/js/sitewide1.js). Its behavior
+    is exercised by a dependency-free Node test (run here when node is
+    available); the Django-side tests assert the wiring that makes the guard
+    effective: the login form posts to the URL the guard watches, and every
+    page loads the script that carries it.
     """
 
     def test_login_page_wires_up_guard(self):
@@ -460,13 +462,21 @@ class LoginDoubleSubmitGuardTests(TestCase):
         # the login form still posts to the action the guard is scoped to
         self.assertIn('action="/accounts/superlogin/"', content)
 
-    def test_sitewide_js_contains_guard(self):
+    def test_guard_behavior_via_node(self):
         import os
-        path = os.path.join(
+        import shutil
+        import subprocess
+        import unittest
+        if not shutil.which("node"):
+            raise unittest.SkipTest("node not available")
+        test_js = os.path.join(
             os.path.dirname(os.path.dirname(os.path.abspath(__file__))),
-            "static", "js", "sitewide1.js",
+            "static", "js", "tests", "login_guard_test.js",
         )
-        with open(path) as f:
-            js = f.read()
-        self.assertIn("data-login-submitted", js)
-        self.assertIn("/accounts/superlogin/", js)
+        result = subprocess.run(
+            ["node", test_js], capture_output=True, text=True, timeout=30,
+        )
+        self.assertEqual(
+            result.returncode, 0,
+            "guard behavioral tests failed:\n%s\n%s" % (result.stdout, result.stderr),
+        )
--- a/frontend/tests.py
+++ b/frontend/tests.py
@@ -441,3 +441,32 @@ class CampaignRetirementTests(TestCase):
         for work in (self.work, self.b2u_work):
             r = anon_client.get("/work/{}/".format(work.id))
             self.assertEqual(r.status_code, 200)
+
+
+class LoginDoubleSubmitGuardTests(TestCase):
+    """Smoke tests pinning the #1240 double-submit guard in place.
+
+    The guard itself is client-side JS (static/js/sitewide1.js); these tests
+    assert the wiring that makes it effective: the login form posts to the
+    URL the guard watches, and every page loads the script that carries it.
+    """
+
+    def test_login_page_wires_up_guard(self):
+        r = Client().get("/accounts/superlogin/")
+        self.assertEqual(r.status_code, 200)
+        content = r.content.decode()
+        # base.html loads the sitewide script that contains the guard
+        self.assertIn("/static/js/sitewide1.js", content)
+        # the login form still posts to the action the guard is scoped to
+        self.assertIn('action="/accounts/superlogin/"', content)
+
+    def test_sitewide_js_contains_guard(self):
+        import os
+        path = os.path.join(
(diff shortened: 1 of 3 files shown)


rdhyee:
**[CC]** Review history — 4 rounds of Codex adversarial review, every substantive round found real defects (final verdict below):

- **Round 1 (NOT-LGTM)**: the 8s timed unlock could itself recreate the 403 (unlock while a slow login navigation is pending → resubmit with stale token); per-form-node lock defeated by a reopened lightbox's fresh node; bfcache restore left a stuck, stale form; smoke tests didn't exercise behavior; substring action matching. → Fixed: lock made global, timed unlock removed, pageshow reload, Node behavioral tests, pathname matching.
- **Round 2 (NOT-LGTM)**: a later document listener cancelling the submission left the lock stuck; target `stopPropagation()` bypassed bubble-phase acquisition; blocked submissions still ran downstream handlers; blocked fresh lightbox forms failed silently; origin unchecked. → Fixed: capture-phase acquisition with deferred rollback, `stopImmediatePropagation`, blocked-button disable, origin check.
- **Round 3 (NOT-LGTM)**: the deferred rollback itself was unsound — `defaultPrevented` at timer time can't distinguish "cancelled" from "cancelled and replaced by `form.submit()`", so rollback could unlock mid-POST (the original bug); window-capture listeners could bypass a document-level guard; `stopPropagation` fallback weaker than documented; `event.submitter` ignored. → Fixed: **fail-closed** lock (no rollback; Codex's recommended design for this codebase), guard moved to window capture, unconditional `stopImmediatePropagation`, submitter-aware button disable.
- **Round 4 (LGTM)**: all round-3 fixes verified faithful; zero remaining should-fix/blocker findings. Accepted limitations (documented in code): a hypothetical future handler that cancels-without-navigation leaves the form locked until reload; native `form.submit()` from an extension's isolated world is invisible to any event-based guard (the observed password-manager behavior does fire the event and is covered).

Codex round-4 verdict, verbatim: *"No remaining should-fix or blocker findings for this codebase. … Verdict: LGTM"* (reviewed PR head `78ae8aa3`; 16/16 Node behavioral tests; JS syntax checks pass).

CC (author) verification: Django suite `libraryauth + frontend + utils` at the pre-existing 3F+2E environmental baseline on every commit — zero net-new failures. Live root-cause evidence in #1240.

https://claude.ai/code/session_01EqzfARuryr6m6jzGscgAY6

PR #1245, opened 2026-09-02
Pre-removal set-a-password email for Google-only accounts (#1237)

rdhyee:
## Summary

- #1237 (removing "Sign in with Google") is gated on real data: a prod query (2026-08-31, read-only, counts only -- see [#1237 comment](https://github.com/Gluejar/regluit/pull/1237#issuecomment-5486695975)) found **8,457 Google-auth accounts with no usable password** (149 active in the last 90 days), all with an email address on file. Removing the button without giving them a way to set a password first would strand those accounts.
- Adds `libraryauth/management/commands/email_google_users_set_password.py`: finds active, no-usable-password users with a `google-oauth2` or legacy `google` social-auth row, and can email each one a link to unglue.it's **existing** "forgot your password?" flow (`libraryauth_password_reset` / `SocialAwarePasswordResetForm` already accepts no-password accounts -- no new view or token logic needed).
- Email copy (`registration/google_set_password_email*.txt`) is **DRAFT COPY** -- Raymond will rewrite it before this ships. A test guards that the marker stays until he does.

## Safety (three independent layers, none disabled this session)

1. **Dry-run by default** -- no `--send`, nothing is sent.
2. **`--send` alone is still a no-op** -- it also requires the `REGLUIT_ALLOW_SET_PASSWORD_EMAIL_SEND` environment variable set to a truthy value. Never set in this session.
3. **`EMAIL_SAFE_MODE` (#1239) is the net underneath both** -- on any non-prod box with it enabled, real sends get allowlist-redirected regardless of what this command does. That's what makes rehearsing `--send` against a real (prod-copy) staging DB safe, once someone actually wants to.

**No sending of any email, no deploys, no prod DB access** -- this session's boundary, held throughout.

## Test plan

- [x] 16 new tests in `libraryauth/tests.py`: the candidate-selection query (both providers, usable-password exclusion, no-email exclusion, inactive exclusion, dedup across two social-auth rows for one user) and the command itself (dry-run, all 4 combinations of `--send`/env-var, `--limit`, and confirming a usable-password user is never emailed even under `--send`).
- [x] `manage.py test libraryauth frontend` -- 68 tests, 3 failures + 2 errors, identical to the documented pre-existing baseline (RhPageTests ×3 CSRF-in-test-client, `registration_complete` NoReverseMatch, `frontend.views` import-order flake) -- zero net-new failures, all 16 new tests pass. Python 3.12 venv + disposable Docker MySQL 8, torn down after the run.
- [ ] Not done here, deliberately: staging rehearsal of `--send` (needs `EMAIL_SAFE_MODE` + the env var set on a box, and RY's rewritten copy first); persistent "already emailed" tracking across runs (noted as a gap in the command's docstring -- `--limit` covers rehearsal scale, a real send needs that or batching first).

https://claude.ai/code/session_01P3dqRP4kc4nFKDQs5FC7gJ

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,11 @@ settings/log_config.py
 settings/prod.py
 settings/aws.py
 
+# Throwaway local test settings (disposable Docker MySQL, dummy creds only)
+# left by CC sessions running the test suite locally — never commit.
+settings/cc_test_csrf.py
+settings/cc_test_local.py
+
 # AWS credential symlinks / key files
 awskeys.py
 iam_keys
--- a/frontend/templates/registration/google_set_password_email.txt
+++ b/frontend/templates/registration/google_set_password_email.txt
@@ -1,4 +1,5 @@
 {# DRAFT COPY -- Raymond will rewrite this before it ships. See #1237. #}
+{% autoescape off %}
 Greetings {% if user.get_full_name %}{{ user.get_full_name }}{% else %}{{ user }}{% endif %},
 
 You're getting this email because your {{ site_name }} account signs in with
@@ -20,3 +21,4 @@ Questions? Just reply to this email.
 
 Best,
 {{ site_name }}
+{% endautoescape %}
--- a/frontend/templates/registration/google_set_password_email.txt
+++ b/frontend/templates/registration/google_set_password_email.txt
@@ -1,0 +1,22 @@
+{# DRAFT COPY -- Raymond will rewrite this before it ships. See #1237. #}
+Greetings {% if user.get_full_name %}{{ user.get_full_name }}{% else %}{{ user }}{% endif %},
+
+You're getting this email because your {{ site_name }} account signs in with
+Google, and we're retiring the "Sign in with Google" button. Your account
+and everything in it (pledges, library, wishlist) is staying exactly as it
+is -- you just need a password so you can still get in afterward.
+
+Set one here:
+
+{{ reset_url }}
+
+Your username, in case you've forgotten: {{ user.username }}
+
+If you don't do anything, your account isn't deleted or locked -- but you
+won't be able to log in until you set a password using the link above (or
+the "Forgot your password?" link on the sign-in page, any time).
+
+Questions? Just reply to this email.
+
+Best,
+{{ site_name }}
--- a/frontend/templates/registration/google_set_password_email_subject.txt
+++ b/frontend/templates/registration/google_set_password_email_subject.txt
@@ -1,1 +1,1 @@
-{# DRAFT COPY -- Raymond will rewrite before this ships. See #1237. #}Set a password for your {{ site_name }} account
+{# DRAFT COPY -- Raymond will rewrite before this ships. See #1237. #}{% autoescape off %}Set a password for your {{ site_name }} account{% endautoescape %}
--- a/frontend/templates/registration/google_set_password_email_subject.txt
+++ b/frontend/templates/registration/google_set_password_email_subject.txt
@@ -1,0 +1,1 @@
+{# DRAFT COPY -- Raymond will rewrite before this ships. See #1237. #}Set a password for your {{ site_name }} account
--- a/libraryauth/management/commands/email_google_users_set_password.py
+++ b/libraryauth/management/commands/email_google_users_set_password.py
@@ -1,0 +1,190 @@
+"""Pre-removal "set a password" email for Google-only accounts (#1237).
+
+Context: #1237 removes the "Sign in with Google" button. A prod query
+(2026-08-31, read-only, counts only -- see PR #1237 comment) found 8,457
+Google-auth accounts with no usable password (149 active in the last 90
+days), all with an email address on file. Removing the button without first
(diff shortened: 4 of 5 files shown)


rdhyee:
## Codex round 1: NEEDS CHANGES, 4 real findings — fixed

Ran an adversarial review (not a rubber stamp) against the initial diff. All four findings reproduced independently before fixing:

1. **High** — negative `--limit` nearly disables the limit (`list[:-1]` is "all but the last one" in Python, not a rejection). `--limit -1` against 8,457 candidates processed 8,456 of them.
2. **High** — no restart-safety (a rerun after interruption re-emails everyone) and per-message SMTP connection overhead.
3. **High** — systemic failures (unreachable mail server, a broken template) were indistinguishable from isolated bad-address failures — the command would grind through the whole list and still print "SUCCESS," and `send_mail()`'s return value was never checked.
4. **Medium** — plain-text templates didn't disable autoescape; a name like "D&Arcy" would render as "D&amp;Arcy" in the actual email.

**All four fixed** in `01fa159e`: an argparse validator rejects negative `--limit`/`--after-id`/`--max-consecutive-failures`; `--after-id` adds a manual restart checkpoint; the mail connection opens once (+ one preflight render) before the send loop so a systemic failure aborts before contacting anyone, a consecutive-failure circuit breaker (default 5) stops a doomed run early, and any unresolved failure now exits non-zero (`CommandError`); both templates got `{% autoescape off %}`. 12 new tests added (28 total in the two new test classes).

Verified: `manage.py test libraryauth frontend` — 80 tests, 3 failures + 2 errors, identical to the documented pre-existing baseline — zero net-new failures.

Codex round 2 (verifying the fixes, not just re-reading the diff) is running now; will post its verdict here too.

rdhyee:
## Codex round 2: NEEDS CHANGES, 4 more findings (verifying the round-1 fixes, not just re-reading) — fixed

1. **High** — `--after-id` is unsafe after mixed success/failure. The loop deliberately continues past isolated failures (that was round 1's own explicit ask), so a failed id followed by a successful one means a naive "resume after the successful id" silently skips the failed one forever — an id cursor can't represent holes.
2. **High** — negative `--limit`/`--after-id`/`--max-consecutive-failures` still got through when this command is invoked programmatically via `call_command(..., limit=-1)`. Django only runs argparse's `type=` conversion on string CLI-style args, not on already-typed keyword arguments — round 1's validator only covered the CLI path.
3. **Medium** — the circuit-breaker summary said e.g. "2 of 2 failed" without mentioning the 3 candidates never attempted at all — reads as "that's everyone."
4. **Medium** — the `with get_connection() as connection:` form let a `close()` exception propagate past the `return`, discarding the sent/failed counts even after real, successful deliveries.
5. **Low** — `--send --limit 0` still opened a mail connection for nothing.

**Fixed in `6b2ab90c`**: didn't build a full delivery ledger (still explicitly out of scope for this pass) — instead made `--after-id` honest about what it guarantees (only "not re-attempted," never "already succeeded") and now print every failed id explicitly so it can be retried deliberately; added a `handle()`-level re-check of all three int options that covers the `call_command(**kwargs)` path; summary now reports attempted/selected/never-attempted counts; connection open/close restructured to explicit `try/finally` so a `close()` failure is logged as a warning instead of losing the accounting; empty-candidate `--limit 0` now returns before touching the mail backend at all.

5 new tests (85 total). `manage.py test libraryauth frontend` — 85 tests, 3 failures + 2 errors, identical baseline, zero net-new failures.

Round 3 (asking Codex for an explicit ship/no-ship verdict, including on the `--after-id` design tradeoff) is running now.

rdhyee:
## Codex round 4 (final): CC+Codex+LGTM

> No blocking findings remain. **Yes—PR #1245 is ready to ship for its explicitly supervised/rehearsal scope.**
>
> Verified at current remote head `22deed65`:
> - Retry instructions correctly use `--after-id <N-1> --limit 1 --verbose-list`, with filtering before limiting and a passing regression test.
> - Failure summaries contain IDs only by default; addresses require `--verbose-list`.
> - `_require_non_negative()` rejects floats, booleans, negative values, and other non-integers for all three integer-shaped options.
> - All 28 command tests passed, including the four new tests.
> - Diff and syntax checks passed.
>
> The deliberate absence of a durable ledger remains appropriate only for the stated supervised scope—not an unattended bulk campaign.

Four adversarial rounds total (round 1: 4 findings, round 2: 5 findings, round 3: 3 findings including one blocking, round 4: clean). Every finding across all four rounds was reproduced independently before being fixed — see the round-by-round comments above for detail. Labeled `CC+Codex+LGTM`.

**Still needed before any real send**: Raymond's review/merge decision, his rewrite of the DRAFT COPY in the email templates, and a decision (with Eric) on which rollout option to use — see `ERIC_CALL_2026-09-03.md` in the repo (this session's other deliverable) for that discussion. No email has been sent, no deploy has happened, and `--send` was never engaged against a real backend at any point in this work.

rdhyee:
Closing — this was the migration step for #1237, which is closed. Google login stays, so no set-a-password email is needed.

PR #1247, opened 2026-09-03
Pass transaction.user through to make_account for donation tokens (#1125)

### LINKED ISSUE #1125: Logged-in donations create anonymous Stripe Customer due to token-presence check

rdhyee:
## Summary

Fixes #1125: a logged-in donor's Stripe Customer was created as `"anonymous user"` because `Pay.__init__` used token-presence alone to decide "user is anonymous" — a fresh donation always carries a Stripe.js token (no `Account` exists yet), so `transaction.user` was silently dropped even for authenticated donors.

- `payment/stripelib.py`: `Pay.__init__` now passes `user=transaction.user` through to `make_account` alongside the token (`transaction.user` is `None` for genuinely anonymous transactions, so that path is unchanged). `make_account`'s existing `if user:` branch already labels the Customer correctly and links the resulting `Account` back to `user.profile` — this also fixes the issue's secondary complaint that saved-card reuse never worked for logged-in donors.
- `payment/tests.py`: new `AnonymousStripeCustomerBugTest`, using the repo's mocked-`StripeClient` pattern (see `StripeWebhookCourtesyEmailTest`) — no live Stripe calls:
  - `test_logged_in_donation_creates_identified_customer` — mirrors `FundView.form_valid`'s plain-donation path; asserts `create_customer` gets `description=user.username`/`email=user.email` and `Account.user == user`.
  - `test_truly_anonymous_donation_still_anonymous` — same shape with `transaction.user = None`; asserts the pre-existing anonymous behavior is preserved.

## Test plan
- [x] Reproduced against the pre-fix code (`git stash` on `stripelib.py`): `test_logged_in_donation_creates_identified_customer` fails exactly as expected (`'anonymous user' != '<username>'`); the anonymous test is unaffected.
- [x] Restored the fix — both new tests pass.
- [x] `manage.py test payment frontend` (47 tests) holds at the pre-existing baseline (3 failures + 1 error + 2 skipped) — re-verified the identical baseline against unmodified `master`. The 1 error (`test_paymentmanager_charge`) makes a real, unmocked Stripe API call and fails on `AuthenticationError` in this sandboxed environment (no real Stripe test-mode key available); reproduced identically on master, unrelated to this change.
- [x] Zero net-new failures.

— rdhyee + Claude

https://claude.ai/code/session_01MHjXjrfha4X6TTSzY5obUZ

### DIFF
--- a/payment/stripelib.py
+++ b/payment/stripelib.py
@@ -650,10 +650,17 @@ class Processor(baseprocessor.Processor):
 
             # ASSUMPTION:  a user has any given moment one and only one active payment Account
             if token:
-                # user is anonymous
+                # A token means we need to create a new Stripe Customer -- but
+                # that doesn't mean the user is anonymous (#1125): a logged-in
+                # user entering fresh card details for a one-off donation also
+                # arrives with a token. Pass transaction.user through (it's
+                # None for genuinely anonymous transactions) so make_account
+                # labels the Stripe Customer correctly and links the new
+                # Account back to the user's profile instead of always
+                # falling into the anonymous branch.
                 try:
                     account = transaction.get_payment_class().make_account(
-                        token=token, email=transaction.receipt)
+                        user=transaction.user, token=token, email=transaction.receipt)
                 except StripelibError as e:
                     self.errorMessage = str(e)
                     return
--- a/payment/tests.py
+++ b/payment/tests.py
@@ -529,20 +529,37 @@ class AnonymousStripeCustomerBugTest(TestCase):
         user = User.objects.create_user(
             'donor_1125', 'donor_1125@example.org', 'payment_test')
         t = self._make_transaction(user)
+        # A Stripe.js-token-shaped string (as FundView actually passes --
+        # the raw dict-of-card-fields shape is only used by the legacy
+        # live-API test elsewhere in this file), so the
+        # create_customer/create_charge assertions below mean something.
+        # Not a real credential -- StripeClient is fully mocked below.
+        card_ref = 'card-ref-test-1125'
 
         with mock.patch.object(stripelib, "StripeClient") as sc:
             sc.return_value.create_customer.return_value = self._fake_customer()
             sc.return_value.create_charge.return_value.id = 'ch_test_1125'
             pm = PaymentManager()
-            pm.charge(t, token={
-                "number": "4242424242424242", "exp_month": 1, "exp_year": 2030})
+            pm.charge(t, token=card_ref)
 
         _, kwargs = sc.return_value.create_customer.call_args
+        self.assertEqual(kwargs.get('card'), card_ref)
         self.assertEqual(kwargs.get('description'), user.username)
         self.assertEqual(kwargs.get('email'), user.email)
 
         account = Account.objects.get(account_id='cus_test_1125')
         self.assertEqual(account.user, user)
+
+        # Codex round-1 (#1247): a failure after Customer creation must not
+        # go unnoticed -- pin down that the charge actually landed on the
+        # Customer this call just created, and that the transaction is
+        # left fully complete.
+        _, charge_kwargs = sc.return_value.create_charge.call_args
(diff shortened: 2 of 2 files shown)


rdhyee:
**Codex review, 2 rounds — CC+Codex+LGTM**

**Round 1**: fix confirmed correct, tests confirmed to genuinely catch #1125. One finding (out of scope for this PR): `Pay.__init__` hands `Execute` the just-created `account.account_id` via `transaction.preapproval_key`, but `Execute` prefers a fresh `transaction.user.profile.account` DB lookup over it — a narrow race (two concurrent submissions for the same user) could charge a different, concurrently-created Customer. Filed as [#1249](https://github.com/Gluejar/regluit/issues/1249) rather than folded in here, since it's a pre-existing `Execute` account-resolution characteristic shared with delayed/queued campaign executions, not something this fix introduces. Took the round's cheap, in-scope coverage suggestions: token-shaped test value, assert `create_customer` got `card=<token>`, assert `create_charge` targeted the newly-created Customer, assert the transaction actually reaches `COMPLETE` with the right `pay_key`.

**Round 2 verdict: ship, no blocking issues.**
- Confirmed deferring the concurrency finding to #1249 is the right scope call — a correct fix requires distinguishing immediate token-backed charges from queued charges that may intentionally want the user's newest card, which doesn't belong in this narrowly-scoped fix.
- Confirmed the hardened test is materially stronger (token shape, card/description/email assertions, Account ownership, charge-target assertion, completion + `pay_key`) and **reran it against the pre-fix `stripelib.py` independently** — confirmed it fails on `"anonymous user" != "donor_1125"`, proving it genuinely detects the regression.
- Confirmed the implementation itself is correct: `user=transaction.user` activates `make_account`'s existing identified-user behavior; `user=None` preserves the anonymous path unchanged.
- Only remaining note was optional polish (`assert_called_once()` on the mocks) — applied.

— rdhyee + Claude

Issue #1253, opened 2026-09-09
Optimize /free/ facet-browse query: missing composite index causes temp-table+filesort, hit ~86K times/day


rdhyee:
## Summary

The free-ebook facet-browsing page (`/free/<facet>/`, `frontend.views.FacetedView`) runs a
`DISTINCT` query joining `core_work` → `core_edition` → `core_ebook` (+ `core_identifier` for
DOAB facets), filtered by `is_free` + one or two facet columns + `language`, ordered by
`featured DESC, created DESC`, paginated via `LIMIT`/`OFFSET`. No index covers this
filter+sort combination, so MySQL falls back to an index-merge intersect, a temporary table,
and a filesort for every request.

This is very likely a real contributor to the DB-CPU-pegged-at-99.9% "slow but up" episodes
from 2026-08-29–09-02 (see [[Unglue.it DB Latency Evidence 2026-09]] internally — happy to
paste the relevant section here if useful).

## Evidence

**1. Caught live on test's new slow-query log** (2026-09-09, under 2 hours of light traffic):

| Query time | Rows sent | Rows examined | Filter |
|---|---|---|---|
| 4.72s | 21 | 54,787 | format=pdf, rights=CC BY-NC-ND, language=fr, OFFSET 100 |
| 1.10s | 21 | 14,120 | DOAB, format=epub, language=pt, OFFSET 60 |
| 2.50s | 21 | 35,693 | DOAB, rights=CC BY-NC-SA, language=de, OFFSET 20 |

Rows examined vs. rows returned: 600–1,700×.

**2. `EXPLAIN FORMAT=TREE`, same query shape, both test and production** (production plan
below — test's is identical in shape):

```
-> Limit/Offset: 21/60 row(s)
    -> Sort row IDs: core_work.featured DESC, core_work.created DESC, limit input to 81 row(s) per chunk
        -> Table scan on <temporary>  (cost=3717..3720 rows=113)
            -> Temporary table with deduplication  (cost=3716..3716 rows=113)
                -> Nested loop inner join  (cost=3705 rows=113)
                    -> Nested loop inner join  (cost=3311 rows=453)
                        -> Nested loop inner join  (cost=3178 rows=348)
                            -> Filter: ((core_work.is_free = 1) and (core_work.`language` = 'pt'))  (cost=1000 rows=2581)
                                -> Intersect rows sorted by row ID  (cost=1000 rows=2582)
[...]

PR #1254, opened 2026-09-10
robots.txt: keep meta-webindexer out of /free/ and other listing pages (refs #1253)

rdhyee:
Refs #1253 — the cheap, no-database half of the `/free/` slowdown fix. The index half is a separate PR.

## What this changes
`meta-webindexer` gets its own `robots.txt` group. It keeps the crawler out of `/free/` and the other expensive listing and feed pages (the same set ClaudeBot already avoids), while individual work pages stay indexable.

## Why
On 2026-09-09, `/free/...` got 86,287 requests. 48,318 of them were from `meta-webindexer`, which had no group of its own and so fell through to `User-agent: *`, which doesn't exclude `/free/`.

## Why not block it outright
Meta's crawler documentation describes it as navigating "the web to improve Meta AI search result quality for users" — a search crawler. This file's stated policy leaves search crawlers able to index the site and blocks only model-training crawlers. So this uses the ClaudeBot-style exclusion (restate the baseline, add the listing paths) rather than `Disallow: /`. No `Crawl-delay`, because Meta doesn't document support for it.

## Caveat
This helps only if the crawler honors `robots.txt`. Meta's page doesn't say either way. Worth rechecking the `/free/` user-agent counts in the Apache log a few days after deploy.

## Tests
- `RobotsTxtTests` asserts the new group's rules and adds it to the exhaustive group-set check: 4/4 pass.
- Negative check: with master's `robots.txt` swapped in, the updated test fails (`'meta-webindexer' not found`).

https://claude.ai/code/session_01DjPGJQCBFvZ1D2fJgJQpe4


rdhyee:
**Codex review, 3 rounds: CC+Codex+LGTM** (at `aaf3a5c1`)

**Round 1 (`bd1341d2`): LGTM, with non-blocking suggestions.**
- Codex confirmed group matching per RFC 9309: `meta-webindexer` matches case-insensitively and doesn't collide with `meta-externalagent`. It also confirmed the change doesn't widen access relative to the `*` group.
- It found that the test didn't pin the exact rules: adding `Allow: /free/`, `Disallow: /work/` or `Crawl-delay: 30` still passed.
- Wording nits: "not blocked outright", "work pages remain crawlable". An older test comment also wrongly claimed that blocking a search crawler sheds no crawl load.
- **Fixed in `f5aacfc4`:** exact Disallow set, empty Allow list, reworded comments.

**Round 2 (`f5aacfc4`): LGTM, with two small points.**
- `other == []` would also reject a legitimate `Sitemap:` line, so the test now checks for `crawl-delay` specifically.
- The duplicate-detection comment was reworded to match what the test actually guarantees.
- **Fixed in `aaf3a5c1`.**

**Round 3 (`aaf3a5c1`): "No findings." LGTM.**

**Verification (local, MySQL 8.0):**
- `RobotsTxtTests`: 4/4 pass.
- Adding `Allow: /free/`, `Disallow: /work/` or `Crawl-delay: 30` to the group fails the test, as intended.
- Adding a `Sitemap:` line passes.
- Master's `robots.txt` fails.

**Caveat, as noted in the description:** this only helps if the crawler honors `robots.txt`. Worth rechecking `/free/` user-agent counts in the Apache log a few days after deploy.

Issue #1257, opened 2026-09-10
Cache the /free/ and /creativecommons/ listing pages (shared cache backend instead of per-process LocMem)


rdhyee:
## Problem

The listing pages served by `FacetedView` — `/free/…` and `/creativecommons/…` — are rendered from scratch on every request. They are among the most expensive pages on the site (see #1253: multi-table `DISTINCT` queries that build a temp table and sort it), and they are also what crawlers hit hardest: `/free/` alone took 41K–111K requests a day in early September 2026, and most of the database load in production's own Performance Insights top-15 over 9/2–9/9 came from these listing queries.

Blocking or throttling individual crawlers in robots.txt (#1173, #1254) helps, but it is whack-a-mole — when `meta-externalagent` was disallowed on 9/9, `meta-webindexer`'s request rate rose within the hour. Caching makes the cost of a repeated listing request small no matter who sends it.

## What exists today

- `settings/common.py` sets `CACHES` to `django.core.cache.backends.locmem.LocMemCache`, and the production settings template in regluit-provisioning does not override it. LocMem is **per process**: each mod_wsgi daemon process has its own cache, nothing is shared, and everything is lost on restart.
- Some template fragments already use `{% cache %}` (`work.html`, `book_panel.html`, `home.html`, `edition_display.html`), and the OPDS / OPDS-JSON feeds cache values via `cache.get`/`cache.set` — all on that per-process LocMem cache.
- `FacetedView` (`frontend/views/__init__.py`) has no caching.

## Proposal (classic best practice, in order)

1. **Shared cache backend.** Move `CACHES` to something all app processes share. Options, cheapest first:
   - Memcached or Redis on the app server itself (no new AWS service; small memory footprint).
   - File-based cache on local disk (zero new services; shared across processes on one box).
   - Not the database cache backend — it would add writes to the database this is trying to relieve.
   - ElastiCache only if we ever run more than one app server.
2. **Cache anonymous listing responses.** Per-URL cache for `/free/…` and `/creativecommons/…` (e.g. `cache_page` or a view-level wrapper) with a short TTL (10–60 minutes; new books arrive nightly, so staleness is cheap). Must only apply to anonymous requests, or cache below the per-user parts of the page — check what the template renders for logged-in users before choosing.
3. **HTTP caching headers** (`Cache-Control: public, max-age=…` for anonymous listing pages), which also makes a CDN in front of the site possible later.
[...]

PR #1255, opened 2026-09-10
core.Work: composite indexes for /free/ facet browsing (refs #1253)

rdhyee:
Refs #1253: the database half of the `/free/` slowdown fix. The `robots.txt` half is #1254.

**Draft: not ready to merge.** The real-data measurement is done (see comments): 4 of 6 test cases got 6×–280× faster, and 2 join-driven cases were unchanged. The migration rehearsal on test is done too (2026-09-10) (see "Merge gates").

## What this changes
Two composite indexes on `core_work`, one per migration:

| Migration | Index |
|---|---|
| `0033_free_facet_lang_index` | `core_work_free_lang_new_idx (is_free, language, featured DESC, created DESC)`, for `?pub_lang=` views |
| `0034_free_facet_index` | `core_work_free_new_idx (is_free, featured DESC, created DESC)`, for views without a language |

Each migration uses `SeparateDatabaseAndState`:
- **Database side:** `RunSQL` creates the index with `ALGORITHM=INPLACE LOCK=NONE`, so MySQL fails instead of falling back to a table copy that blocks writes. It runs under a 10-second session `lock_wait_timeout`, so the migration fails fast rather than queueing an exclusive metadata lock behind a long-running query. Reverse SQL drops the index the same way.
- **State side:** a matching `AddIndex`, so `Work.Meta.indexes` and migration state agree.

## Why
Every `/free/` facet query starts from `Work.objects.filter(is_free=True)`, optionally narrows by language, and by default sorts by `-featured, -created`. Today, `EXPLAIN` on test and production shows MySQL intersecting the separate `is_free` and `language` indexes, then building and sorting a temporary table. Test's slow-query log caught three such queries taking 1–5 s and examining 14K–55K rows to return 21. Production served 86,287 `/free/` requests on 2026-09-09.

The *intended* benefit is one range scan in sort order in place of the index intersection. The query also does `DISTINCT` across joined facet tables, which may still force a temporary table. That has to be measured, not assumed.

## Verified (local disposable MySQL 8.0)
- `makemigrations --check`: no changes detected.
- Full `migrate` creates both indexes with the expected columns and sort directions (`is_free A, language A, featured D, created D` and `is_free A, featured D, created D`).
- `migrate core 0032` removes both; re-applying recreates them.

## Merge gates (not done yet)
[...]

PR #1258, opened 2026-09-10
robots.txt: Crawl-delay experiment for meta-webindexer (refs #1253)

rdhyee:
## What

Adds `Crawl-delay: 10` to the `meta-webindexer` group in `robots.txt`, as an explicitly labeled **experiment**, and updates `RobotsTxtTests` to expect exactly that value. Follow-up to #1254 (merged) for #1253.

## Why an experiment

`Crawl-delay` is not part of the Robots Exclusion Protocol (RFC 9309 §2.2.4 lets crawlers ignore records it doesn't define), and Meta's crawler documentation says nothing about crawl rate. So there's no documentation to rely on. But production's access logs make this unusually easy to measure:

- On 2026-09-09 at 12:00 UTC, `meta-externalagent` stopped (≈1 hour after Meta re-read the robots.txt that disallowed it), and in the same hour `meta-webindexer` jumped to a **flat ~7,200 requests/hour** (≈2/s), where it has stayed since.
- A 10-second delay would cap it at 360/hour. A fall to that range after Meta's next robots.txt fetch would support keeping the line (support, not proof — other changes could also move the rate); a rate that stays near 7,200/hour means it should be removed.

## A related finding, recorded in the comment

Both Meta crawlers crawled `/feedback/` — excluded only by the `*` group — tens of thousands of times a day (`meta-externalagent` 9/1–9/8, `meta-webindexer` 9/9–9/10), while `meta-externalagent` stopped within about an hour of Meta re-fetching robots.txt once it had its own `Disallow: /` group. So a named group is the mechanism with evidence behind it. It has to carry every rule itself anyway, since named groups don't inherit `*` (it already does).

## How to read the result

After deploy, Meta re-reads robots.txt via `facebookexternalhit` (a burst around 11:00 UTC daily in recent logs). Then, per UTC hour for `meta-webindexer`:

| Observation | Meaning |
|---|---|
| `/free/` and other listing requests → ~0 | #1254's `Disallow` honored |
| Total ≤ ~360/hour | Supports `Crawl-delay` being honored — keep |
| Total still ~7,200/hour | `Crawl-delay` not reducing the rate — delete the line |
| In between | Ambiguous — look at the path mix before concluding |

Revert = delete the `Crawl-delay: 10` line and set the test back to expecting none.

## Testing

- `frontend.tests.RobotsTxtTests` passes.
- Mutation checks: removing the line, changing it to `5`, and duplicating it each make the test fail.

## Related

#1253, #1254, #1257 (caching the listing pages), #1255.

https://claude.ai/code/session_01DjPGJQCBFvZ1D2fJgJQpe4


rdhyee:
**[rbotyee+CC+Codex]** Review status: CC + Codex LGTM at `a2454196`. Not deployed.

**Codex round 1 (`698e0d66`) — VERDICT: LGTM**, two non-blocking wording findings:
- Placement is valid: RFC 9309 §2.2.4 says extension records must not interfere with standard records or end a group; Google's parser ignores `Crawl-delay` without changing group state; Python's `urllib.robotparser` records the delay and keeps the following `Disallow` lines in the same entry. Codex checked access decisions for 12 agents × 15 paths before/after: unchanged, except `meta-webindexer`'s delay is now 10.
- The test requires exactly `["10"]`: missing, wrong, duplicated, empty, commented-out and misplaced delays all fail; an extra `Sitemap` line passes.
- Findings: the comment overgeneralized from `meta-externalagent` to "Meta's crawlers" and said "for weeks"; the experiment note implied a rate drop would prove support.

**Fix (`a2454196`, comment-only):** the named-group observation now names `meta-externalagent` only; the experiment is framed as an assessment with the baseline start (2026-09-09 12:00 UTC).

**Codex round 2 (`a2454196`) — VERDICT: LGTM, no findings.** Both findings resolved; non-comment robots.txt lines and `frontend/tests.py` identical to round 1.

**CC checks:** `frontend.tests.RobotsTxtTests` passes (4 tests); mutation checks — line removed, set to `5`, duplicated — each fail.

https://claude.ai/code/session_01DjPGJQCBFvZ1D2fJgJQpe4

Issue #1261, opened 2026-09-18
Every page mints a unique /feedback/ and sign-in URL — 694k + 176k distinct URLs crawled in one day


rdhyee:
Every page on unglue.it mints **two unique, uncacheable URLs** that exist only for crawlers to follow: a `/feedback/?page=<this page's URL>` link and a Google sign-in link carrying `?next=<this page's URL>`. With roughly two million crawlable pages, that is a URL space of the same size — and both destinations are expensive.

This is what took the site down four times on 2026-09-17 (115 minutes; `EbookFoundation/security-private#32`).

## The measurement that makes the case

From the production access log for 2026-09-17:

| | Requests | **Distinct URLs** |
|---|---|---|
| Whole site | 2,260,927 | 1,866,227 |
| `/feedback/…` | 702,435 | **693,968** |
| Google sign-in | 184,125 | **175,983** |
| `/work/…` (for contrast) | 327,040 | 243,147 |

**Essentially every request to these two endpoints is for a URL never requested before.** That has three consequences worth stating separately:

1. **Caching cannot help.** A cache needs repetition, and the repeat rate here is 1.012. No amount of `mod_cache` touches this.
2. **Rate limiting barely helps.** The traffic is spread across thousands of addresses each making a handful of requests — 6,647 addresses in the window of the fourth outage, the busiest making 119.
3. **Blocking does not fix it.** The fourth outage happened *after* both loud sources were blocked. As Eric put it: blocks do not solve the problem and are not worth the maintenance.

The work is generated by the site itself. Whoever crawls it, the bill is the same.

## Why the July fix did not finish the job

`frontend/templatetags/feedback_link.py` (see [#1204](https://github.com/Gluejar/regluit/issues/1204) and `INCIDENT_2026-07-10_crawler_trap_flood.md`) fixed an *infinite* self-referencing space — each level re-encoded the last (`%2F` → `%252F` → …). That fix is correct and should stay.

But it made the space **finite, not small**. One unique feedback URL and one unique sign-in URL per crawlable page is still ~2M of each. The July work bounded the recursion; it did not stop the minting.

## What would actually fix it

Both endpoints put context in the URL only so the destination can know where the user came from. That context does not have to live in the URL.

[...]

PR #1262, opened 2026-09-18
Collapse the /feedback/ URL space to one URL (#1261)

rdhyee:
## For Eric — the decision at a glance

**Roughly:** the feedback link stops being a different URL on every page. The cost falls on the feedback emails sent to `unglueit@ebookfoundation.org`: some will lose the page they came from, and three links lose their subject-line labels.

**Today**, every page links to its own feedback URL:
```
https://unglue.it/feedback/?page=https%3A%2F%2Funglue.it%2Fwork%2F80457%2F
```
On 9/17 there were 693,968 distinct feedback URLs, each costing a full page render. The staff email reads **"Feedback on page https://unglue.it/work/80457/"**.

| | Before | After |
|---|---|---|
| Link on every page | `/feedback/?page=<this page>` | `/feedback/` (one URL, `nofollow`) |
| Old `?page=` URLs (and any query) | full render | cheap **301 redirect** to `/feedback/` |
| How the form knows the originating page | from `?page=` | from the browser's `Referer` header (same-site only) |
| Staff email when that header is present | "Feedback on page …/work/80457/" | same |
| When the header is missing (privacy settings, links clicked in email) | page still known | **"Feedback on page /"** |
| FAQ "no activation email" link | subject "…no activation email" | subject "…https://unglue.it/faq/" |
| FAQ "need support" link | subject "…need support" | **same FAQ subject**, indistinguishable from the one above |
| Privacy page link | subject "…Privacy" | subject "…/privacy/" |

**What we need from you:** is losing those subject labels, and the originating page on some messages, acceptable? If you want the labels back, they can return later as fixed paths such as `/feedback/support/`. They can't come back as `?page=` parameters, because those now redirect.

---

**TL;DR** — Every page on the site linked to `/feedback/` with its own URL tacked on, so crawlers found a different feedback URL on every page: 693,968 of them on 9/17, each costing a full page render. This makes it one URL. Where someone came from now comes from the browser's `Referer` header instead, and the old URLs get a cheap redirect rather than a render. **One small cost:** three FAQ and privacy links lose the triage hint they used to put in the subject line.

Fixes part of #1261. Merge this before #1263, which is rebased on it.

## What changed

`{% feedback_url %}` emits a constant `/feedback/`. The feedback view reads the originating page from a same-site `Referer`, falling back to `/`. Any non-POST request carrying a query string gets a 301 to the bare URL — **that redirect is what collapses the 693,968 URLs already sitting in crawler queues**, rather than merely declining to mint more.

Two things that the change depends on, and would have been broken without:

- The site has sent `<meta name="referrer" content="origin">` since 2015, which truncates even a same-origin `Referer` to `https://unglue.it/`. So the feedback links carry `referrerpolicy="same-origin"` individually. The site-wide policy is deliberately untouched — changing it would have started feeding full internal URLs to Google Analytics, which is the opposite of what the 2015 change was for.
- Six notification templates hand-wrote the old-style link outside the tag. Converted, with a structural test that fails if anyone reintroduces the pattern.

## The cost, so it's a decision rather than a surprise

Three hand-written links — "I never received an activation email", "I need support", and one on the privacy page — used to pre-fill the subject with a triage category. They now produce the same subject as any other feedback from that page, so the two FAQ links are indistinguishable from each other.

Recoverable later with a fragment or three fixed paths like `/feedback/support/`, but **not** with a query parameter, since those now all redirect. Not done here to keep the change small.

## Worth knowing

- The claim above is true of the `/feedback/` route. `/feedback/campaign/<id>/` is a different view, deliberately excluded, and still renders with a query string — it is linked from every work page with a campaign.
- The form and thank-you page are marked never-cache: they carry a CSRF token, a captcha and the signed-in user's email. The 301 is deliberately cacheable.
- Pre-existing and adjacent: a POST to `/feedback/` missing `num1` raises a 500, which lands in the error mail.

## Verified

24 feedback tests against a disposable Docker MySQL. The redirect gate was checked empirically — including that `?&` and `?&&`, which parse to no parameters at all, still redirect rather than rendering. Both structural tests were confirmed non-vacuous by reintroducing the bad pattern and watching them fail.

Not covered: nothing was run against a server, `core`'s tests were not run, and the full `frontend` suite has six errors that are environmental — their setup fires a celery notification with no broker present, and the same six appear on an unmodified checkout.

https://claude.ai/code/session_01Ky3JyVaftSREv1JSSdwzcv


rdhyee:
**[Codex via CC]** Codex (gpt-5.4, low reasoning effort) reviewed this branch twice — once as a design consult before implementation, and once on the full diff.

**Verdict: LGTM.**

> LGTM.
>
> Re-read the full diff. `referrerpolicy="same-origin"` is spelled correctly and placed on all 17 feedback `<a>` elements; it correctly overrides the global policy for those links.
>
> The assertions match the comments, pagination coverage survives the deduplication, and I found no regression in URL handling, POST preservation, or subject cleaning.
>
> Static review; tests not run.

Two things Codex changed about this PR, worth recording because they were not cosmetic:

**1. It rejected my first fix for the referrer problem.** I had changed the site-wide `<meta name="referrer">` from `origin` to `same-origin`, reasoning that it leaks strictly less at the HTTP-header level. Codex's answer:

> **Google Analytics:** Collection continues. **Privacy caveat:** GA defaults `page_referrer` to `document.referrer`, so newly available full internal referrers can be forwarded to Google in analytics payloads. "Less third-party leakage" is therefore true for the HTTP header, **not necessarily overall**.
>
> **I'd ask for the narrower change and accept that.** It solves #1261 with a bounded exception. A global policy change deserves a separate privacy decision, especially because of analytics forwarding and expanded internal logging.

So the PR now carries `referrerpolicy="same-origin"` on the feedback links only, and the 2015 site-wide policy is untouched. This is the better change and I would not have got there alone.

**2. On the earlier revision it confirmed the compatibility questions** that mattered most: old `/feedback/?page=<encoded url>` links still resolve (the explicit parameter takes precedence over Referer); it found no path by which the URL space can regrow; and dropping `takes_context=True` from the tag breaks no call site.

Its remaining note on that pass was that two of my tests duplicated each other — fixed in 79da4b3b.

rdhyee:
**[Codex via CC]** Second Codex pass, on the commits added after the first LGTM (the six notification templates and the structural test).

**Verdict: LGTM.**

> LGTM — no blocking findings in this incremental diff.
>
> **Template inheritance works.** On local Django 5.2.15, all six child templates compiled and rendered with `{% load feedback_link %}` after `{% extends %}`, including without a request. The child's load makes the tag available inside its blocks; the parent needn't load it.
>
> **Email links remain relative:** previously `/feedback/?page=…`, now `/feedback/`. No regression. Absolute email links are worth fixing separately.
>
> **The structural test catches the old pattern, but isn't exhaustive.** False negatives include double-quoted `"feedback"`, different tag whitespace, multiline attributes, literal `/feedback/`, and an earlier unrelated `?page=` on the same line. […] Useful regression guard, but its "every request-derived feedback URL" claim is too broad.

That last point was fair and is fixed in 8de52d97: the guard now also matches `{% url "feedback" %}` and a literal `/feedback/?page=`, and its comment no longer claims more than it does — it says plainly that an attribute split across lines would slip past. Both new spellings were used as canaries to confirm the test actually fails on them.

The relative-link-in-email note is recorded in "Deliberately not doing" rather than fixed here; it predates this PR.

rdhyee:
**[CC]** Round 3: four review findings addressed. The first one changed what this PR actually does, so it is worth spelling out.

**The PR did not do what its title said.** Honouring `?page=` left the existing space fully live — 693,968 URLs are sitting in crawler queues right now, and every one of them kept returning 200 with a full template render. The day after deploy the same fleet replays yesterday's list and the load is identical. `rel="nofollow"` only suppresses *new* discovery, and `robots.txt` has disallowed `/feedback/` for years — ignored by exactly the crawlers that caused the outage.

So `/feedback/` now takes no query parameters at all: a GET or HEAD carrying any query string gets a 301 to the bare URL, and `?page=` is no longer read. Scoped to the `feedback` route by `resolver_match.url_name`, so `ask_rh` at `/feedback/campaign/<id>/` — which calls the same view — is untouched.

**The three hand-written topic links are removed rather than given `nofollow`.** With the redirect in place `?page=Privacy` would just bounce, and they were the only followable paths left into that space. The cost is real and accepted: staff lose the "no activation email" hint in the subject line and get the FAQ page URL instead.

**Never-cache headers** on the form and thank-you page rather than `Vary: Referer` — the response also carries a CSRF token, a one-time captcha and the signed-in user's email, so it must not be shared at all. Deliberately not applied to the 301, which caches and crawlers should remember.

**A bug Codex caught in my own fix**, worth recording because it would have quietly reopened the hole: I first gated the redirect on `request.GET`, which is not the same as "has a query string" —

> `/feedback/?&` and `/feedback/?&&` parse to an empty `QueryDict`, bypassing the redirect and rendering the form. Use `request.META.get('QUERY_STRING', '')` and test those cases for GET and HEAD.

Verified independently before accepting it (`QueryDict('&')` is falsy), fixed, and both spellings are now in the test.

Codex's round-4 verdict on the corrected diff: **LGTM**, with three comments of mine that overclaimed. All three are fixed — most substantively, a test named "a real user following an old link is still attributed" rested on Django's test client re-sending the `Referer` through a redirect. The test client is not evidence about browsers, and the realistic case is a crawler arriving with no `Referer` at all, landing on `/`. The test remains; the claim around it is now the size of the evidence.

**Testing.** 21 feedback tests pass. The full `frontend` suite shows 6 errors in `RhPageTests` and `CampaignRetirementTests` — environmental, not from this change: their `setUp` creates a `Claim`, whose `post_save` fires a notification through celery, and no redis broker was running on the machine I tested on. Confirmed by stashing this branch and running the committed state alone: same errors.

rdhyee:
**[CC]** Housekeeping: this PR's description was wrong between 17:28 and 18:12 UTC — it showed the description of [#1263](https://github.com/Gluejar/regluit/pull/1263), the sign-in half of #1261. Two agents working on the two halves shared a scratchpad directory and both wrote a file called `pr_body.md`; I then uploaded that file three times without checking it. The description is now restored and current. No code was affected — only this text.

Recording it because anyone who read this PR in that window read the wrong thing.

rdhyee:
**[Codex via CC]** Codex's final pass on the complete branch diff, plus the two product questions the reviewer asked me to put to it explicitly.

**Verdict: LGTM — no blocking findings.**

> The early GET/HEAD redirect, exclusion of `ask_rh`, bounded Referer handling, subject cleaning, and non-cacheable form responses fit together correctly. No further code changes requested.

**On 301 vs 302:**

> Confirm 301: the intended canonicalization is permanent, and cached redirects can survive a server change. A 302 communicates temporary relocation.
>
> **Concrete future cost:** After changing the redirect rule, previously cached *exact query URLs* may still redirect until their cache entries expire or are cleared; a new query URL isn't automatically covered by those old entries.

So the cost is bounded to URLs that have already been redirected — a future `/feedback/?topic=x` would be a URL no cache has an entry for.

**On whether to keep the three topic markers**, and on the static-path alternative:

> Three explicitly matched static paths add only three intended destinations. They do **not inherently multiply per originating page**, but "cannot be fanned out by a crawler" is too absolute: crawlers can append arbitrary queries or invent paths. Reject unknown paths, canonicalize queries on the three routes, and avoid reflecting arbitrary input into links.
>
> **Recommendation:** Keep this PR as-is. The hint is a convenience; defer it unless support staff report a meaningful loss. If automatic link-specific hints later matter, three explicit static routes are a reasonable small follow-up.

Codex also ruled out the two cheaper alternatives I asked about: deriving the hint from the `Referer` path cannot distinguish two feedback links on the same FAQ page, and a topic selector on the form would need the user to choose it. The "Deliberately not doing" section now states the option and its limits in those terms rather than my original, too-confident wording.

Codex has reviewed five times at low reasoning effort across this PR: the design before any code, the first diff, the incremental diff, the round-3 fixes (where it found the `?&` gate bug), and this final pass.

rdhyee:
**[CC]** `CC+Codex+LGTM` applied at `06d3fcc6`. Here is exactly what it rests on, so the label is a summary and not a claim.

**CC-side review** (`/code-review`, two passes). The first produced four findings; the second, deeper pass produced seven. Verdict: *"no blocking correctness bug — it does collapse the space and the July guard is intact."* All eleven are addressed. The substantive ones:

- **The PR was inert as first written.** Honouring `?page=` left the existing 693,968 URLs returning a full render each. That is now a 301 — the change that makes the title true.
- **Four templates missed the `referrerpolicy`** (`404.html`, `500.html`, `503.html`, `registration/welcome.html`) because my sweep looked for `{% feedback_url %}` and they used `{% url 'feedback' %}`. Without it a user reporting a 500 is recorded as having been on the home page — a *wrong* attribution, worse than the honest `/` a missing Referer gives.
- **OPTIONS and TRACE still reached a full render** per distinct URL, being CSRF-exempt. Gate widened to everything but POST.
- **Truncation was silent**, so staff got a cut-off URL that looked complete and 404ed. Now marked.
- **Two tests were weaker than their names.** The tripwire missed `{% feedback_url %}?page=…` — now the idiomatic spelling, so the likeliest way to reintroduce this — and `test_feedback_url_tag_without_request_in_context` had gone vacuous.

**Codex** (five passes, low reasoning effort): on the design before any code, and on each diff. Two of its findings changed the PR rather than polishing it — it talked me out of changing the site-wide referrer policy, which would have fed full internal URLs to Analytics; and it found that `request.GET` is not "has a query string", since `?&` and `?&&` parse to an empty `QueryDict` and would have rendered in full, a crawler hole inside the fix for a crawler hole. Its last pass found three false positives in the `referrerpolicy` guard I had *just* written.

**What the label does not cover.** Nothing has run against a server. The `core` app's tests were not run. The full `frontend` suite is 59 tests with 6 errors, all in `RhPageTests`/`CampaignRetirementTests` and all environmental — their `setUp` creates a `Claim` whose `post_save` fires a celery notification, with no redis broker on the test machine; verified by stashing this branch and running the committed state alone. `ask_rh` has no test beyond the new one asserting the redirect does not catch it.

**One decision for @rdhyee and @eshellman**, in the description under *Deliberately not doing*: dropping the three topic markers costs a triage hint in the subject line of support messages sent from the FAQ. Recoverable later by a fragment or three static paths, not by a query parameter.

Not merged — @rdhyee merges.

PR #1263, opened 2026-09-18
Stop every crawlable page minting a unique Google sign-in URL (#1261)

rdhyee:
**TL;DR** — Every page on the site put its own URL inside the header's Sign In link, so crawlers found a different sign-in URL on every page: 175,983 of them on 9/17, each one expensive because serving it ties up a web worker while we wait on Google. This moves that value out of the link, so crawlers see one sign-in URL instead of two million. **One tradeoff for you to accept or reject:** with JavaScript off, signing in from the header now lands you on the home page instead of the page you were reading. It also closes an open redirect that this change would otherwise put on the main path.

Fixes part of #1261. Merge after #1262 (already rebased on it).

## The decision for @eshellman and @rdhyee

**With JavaScript off, blocked, or if `sitewide1.js` fails to load:** the header Sign In link has nothing saying where you came from, so after signing in you land on the home page and have to find your way back. On that link the lightbox means it fails quietly rather than visibly.

Everything else still works without JavaScript — gift redemption, pledge and purchase logins, and every `@login_required` bounce — because those carry the destination in the URL already.

## What changed

The Sign In / Sign Up links keep `?next=` out of the href and hold it in a `data-next` attribute; eight lines of JavaScript put it back on page load. A real browser gets exactly the URL it got before; the HTML we serve holds one sign-in URL.

Also: the login page's Google link now falls back to a constant `/next/` rather than echoing its own URL; `rel="nofollow"` on the five Google links; a `?v=` cache-bust on `sitewide1.js`, since the behaviour now lives there and static files carry no content hash.

**Honest limit:** this removes the fan-out for crawlers that *don't* run JavaScript — the fleet behind the 9/17 outages. A crawler that renders, as Googlebot does, still sees one sign-in URL per page. `rel="nofollow"` has been a hint rather than a directive since 2019.

## Security: an open redirect, fixed here

The `next` cookie is written by JavaScript from a value the URL can carry, and three places consumed it without checking: the `/next/` view, an automatic redirect at the end of registration, and a "go back to where you were" link. An unvalidated redirect there is a phishing aid — the victim starts on unglue.it and ends up elsewhere after what looks like signing in.

It predates this branch, and is fixed here because **this branch makes `/next/` the default hop after signing in**, which would move a dormant bug onto the main path. Specifics are kept out of a public PR until this deploys; available to maintainers on request.

The server view now requires a relative same-site path and falls back to `/`. The other two no longer read the cookie at all — both point at the guarded `/next/`. There is also a check where a URL value first becomes a stored destination, which is defence in depth rather than the boundary.

Related: #1264 — Django can't tell that requests arrive over HTTPS (`SECURE_PROXY_SSL_HEADER` unset, no HSTS). **Read the warning in that issue before acting on it.**

## Worth knowing

- The welcome-page redirect costs one extra hop, in exchange for the destination being checked on the server.
- A visitor who opens the sign-in lightbox, walks away, and later signs in from `/accounts/superlogin/` will be returned to that earlier page. New, and arguably better than the old outcome, but it is cross-flow coupling.
- A crawler that runs JavaScript, or harvests `data-` attributes, can still recover the per-page value.

## Verified

32 new tests against a disposable Docker MySQL: header links identical across pages and unchanged by query strings, the current page still carried, `@login_required` still works, the gift and pledge flows still point Google at their real destinations, and all three redirect sinks plus the write-side check.

Every guard was reverted and the tests watched to fail — a security test that passes either way is worthless. The write-side check was also run against legitimate values (encoded and plain paths, a literal percent) as well as hostile ones, after an earlier version of it rejected real pages whose URLs contain spaces, like `/bypub/all/Oxford University Press`.

Not covered: nothing was run against a server, and no real browser or Google round trip was exercised. Two of the pre-existing test failures on this branch are stale tests that would fail in CI too — a CSRF regex expecting single quotes, and `reverse('registration_complete')` where the route is `django_registration_complete`.

https://claude.ai/code/session_01Ky3JyVaftSREv1JSSdwzcv


rdhyee:
**[Codex via CC]** Two rounds at low reasoning effort — once on the design before any code was written, once on the diff.

**Design round.** Codex endorsed the data-attribute + JS approach over the alternatives: *"Do not use `Referer` or sessions: both add state/ambiguity and can still create crawler-driven work."* On the residual risk: *"`data-next` does not meaningfully recreate the crawl space for normal crawlers. It is not an anchor URL, so ordinary link extraction will not follow it. A JS-capable or deliberately attribute-harvesting crawler can recover it... Treat this as mitigation, not a security boundary."* It also confirmed the `/next/` fallback on the login page is coherent, and that `hijax`'s `href.split("next=")` still sees the same shape after promotion.

**Code round 1** raised three things: the non-JS regression; that the phrase "the same URL on every crawlable page" claimed more than the diff delivers, since `home.html`, `from_pledge.html` and `gift_login.html` still carry `?next=` in Google sign-in hrefs; and that the comments and test docstring should be cut back to match.

The non-JS loss is deliberate and is the thing called out in the PR description for maintainer judgement. The overclaim was real and is fixed in a87abc8b — by narrowing the wording rather than widening the change, since those three templates are reached from four routes rather than from every page.

**Code round 2, on the final diff:** *"Clean. I found no correctness or safety issue in the final diff. The deliberate non-JS tradeoff is clearly documented, the remaining Google links are bounded and load-bearing, and the tests cover the changed URL-space and redirect behavior."*

rdhyee:
**[CC]** Self-review caught a bug my own tests missed, fixed in 9bdf7c59.

The explanatory comments I added used Django's `{# ... #}` syntax spread over several lines. That syntax is single-line only — a multi-line one is not a comment at all, it is literal text. So the comment text was being served inside the `<ul>` of **every page on the site**, which is a mildly absurd outcome for a change whose whole purpose is to make every page cheaper.

The tests passed the entire time, because they asserted on the hrefs and the hrefs were correct. What caught it was rendering a page and reading the HTML. Now `{% comment %}`, plus a regression test asserting the text does not reach the browser.

Worth recording as the lesson: a green test run tells you what you asserted, not what you shipped.

rdhyee:
**[CC]** `/code-review` on the branch. Three findings; one changed the code.

**Fixed (d0f63412) — the login page's secondary links were still fanning out.** I had listed this under "deliberately not doing" on the grounds that those links sit on a single bounded page. The review made a better argument: *"a crawler appends junk to the login page — `/accounts/superlogin/?utm=1`, `?utm=2`, … Each variant yields a distinct `Need an account` href... `registration_form.html` then feeds `request.GET.next` straight into its Google button, producing... a fresh, uncacheable URL on the exact expensive endpoint this PR is about, two hops from the page the new test declares closed."* That is right, and it is one line, so it is now fixed with the same `/next/` fallback, plus a test.

**Noted, not changed — the `/next/` fallback reads a cookie that can be stale.** Open the Sign In lightbox from `/privacy/` (which writes the `next` cookie), close it without signing in, then later go directly to `/accounts/superlogin/` and use Google — you land on `/privacy/` rather than the home page. That is new: the old code echoed the login URL and never consulted the cookie. It is also arguably an improvement, since the old behaviour returned you to a login page reading "you are already logged in as X". Recording it so it is a known property rather than a surprise.

**Noted — the non-JavaScript loss**, which the review flagged as *"a deliberate, documented tradeoff rather than an oversight; flagging it so it is a decision rather than a surprise."* Exactly the intent: it is the one thing in this PR that needs a human yes.

**Checked and clear**, per the review: the promotion runs before the delegated `a.hijax` click handler and that handler reads `href` at click time, so the lightbox load and the `next`-cookie push both see the promoted URL; `auth_next`'s `quote(..., safe='')` output cannot break an HTML attribute; the promotion is idempotent and picks `?` vs `&` correctly; an empty `data-next` is skipped rather than leaving a dangling `next=`; `sitewide1.js` sits outside any `{% block %}` so `registration_base.html` cannot drop it; and the gift test reproduces the real "established giftee" branch of `receive_gift`.

rdhyee:
**[Codex via CC]** Round 3, on the two commits that landed after its round-2 pass (the comment-leak fix and the login-page fan-out fix):

> Clean.
>
> - `{% comment %}` correctly fixes the multiline-comment leak.
> - The login-page secondary links now use the constant `/next/` fallback, preserving explicit `?next=` destinations and closing the two-hop fan-out.
> - No apparent regressions elsewhere in the branch; unrelated flows remain unchanged.
> - Targeted tests could not run because Python 3.9.11 is unavailable in this checkout.

On that last line: Codex could not run the suite in its own sandbox. I ran it — `frontend` + `libraryauth`, 65 tests, with the same 3 failures and 1 error as a clean `origin/master` baseline run on the same database (`RhPageTests.test_rh/test_staff/test_user`, `TestLibraryAuth.test_registration`), all pre-existing and unrelated to this branch.

Both reviewers are now clean on the final diff.

rdhyee:
**[CC]** Second `/code-review` pass found something that changes this PR's weight. Four findings; three fixed in a8677ea2, one accepted.

> **Edited.** This comment originally spelled out how to exercise the open redirect described below. The repository is public and the fix is not deployed yet, so that detail has been removed until it is. Maintainers can get it on request; the reproduction is in the tests on this branch.

### The important one: an open redirect, now fixed

`frontend/views/next()` redirected to a destination held in a cookie that client-side JavaScript writes, without validating it. An unvalidated redirect there is an open redirect, and a phishing primitive: the victim starts on unglue.it and lands elsewhere after what looks like signing in.

**I confirmed it rather than reasoning about it** — reverting the guard makes the new tests fail, showing the redirect leaving the site.

It predates this branch. But **this branch makes `/next/` the default post-auth hop from the login page**, where the old code echoed the login URL and never touched the cookie — so shipping as-was would have moved a dormant bug into a load-bearing position. Fixed here rather than deferred, using `url_has_allowed_host_and_scheme` against the request's own host and scheme: the same guard Django's own login and logout views use, and the same one `libraryauth/tests.py` already documents for logout. A rejected value is cleared rather than left to be retried. Six new tests.

### Also fixed

- **`rel="nofollow"`** on the login page's "Forgot your password" / "Need an account" links, which d0f63412 fixed for URL fan-out but left followable.
- **Cache-busting on `/static/js/sitewide1.js`.** This is the failure mode a real user actually hits: the "come back here" behaviour now lives in that file rather than in the HTML, and static files are served by plain `StaticFilesStorage` with no content hash — so a returning visitor holding a cached copy would get the new markup with JS that does not know to promote `data-next`, and would silently lose the behaviour until their cache expired. Now `?v=1261`, with a test. A hashed filename (`ManifestStaticFilesStorage`) would be stronger and is a reasonable follow-up, but it is a site-wide change and does not belong in this PR.

### Accepted, not fixed

A click on the header Sign In landing in the window before `document.ready` — or any failure to load jQuery — navigates to the bare URL and loses `next`, with the hijax cookie fallback failing at the same time for the same reason. Narrow, and distinct from the documented JavaScript-off tradeoff: that one is a choice, this one is a JS-on user silently getting the JS-off outcome. Recording it as an accepted risk.

### One reviewer finding I am pushing back on

Codex's fourth round flagged `test_sitewide_js_is_cache_busted` as self-contradictory, on the grounds that `?v=1261` still contains the substring `src="/static/js/sitewide1.js"`. It does not: the asserted string ends in a closing double quote, which in the versioned markup falls after `?v=1261`. Checked both ways — the needle is absent from the versioned tag (so `assertNotIn` passes) and present in the unversioned one (so the test fails if the cache-bust is ever removed, which is the point). The test passes in the suite. Everything else in that round came back clean.

rdhyee:
**[CC]** The open redirect was wider than the first fix covered. Three consumers of the `next` cookie, not one — all now closed, on head `d23d8965`.

I had reported the server-side guard as closing it. That was wrong, and the way it was wrong is worth recording: I fixed the sink I had been looking at, then reasoned about the fix rather than re-asking the original question, which was *who reads this cookie.* Two reviewers asked it properly and found the other two.

| # | Consumer | Was | Now |
|---|---|---|---|
| 1 | `frontend/views/next()` | redirected to the cookie unvalidated | validates against the request's own host and scheme, falls back to `/`, clears the cookie |
| 2 | `registration_base.html` | `window.location.replace(saved_next)` — automatic, never reached the server | navigates to the constant `/next/` |
| 3 | `registration_complete.html` | `<a href="{{ request.COOKIES.next\|urldecode }}">` — user-clickable | links to `/next/` |

Found by the Claude-side reviewer (#2) and by Codex (#3). Neither was reachable from the guard I had written, because neither goes through the server.

**On #3 I checked reachability rather than assuming**, since a dead template would change the severity: the live route `django_registration_complete` renders `django_registration/registration_complete.html`, which is a one-line `{% extends %}` of the template holding the sink. It is live, and the test goes through that real URL.

**Everything is falsified.** Each guard reverted, tests watched to fail. The write-side check was additionally run against real values in node, because that class of check breaks legitimate input silently and this PR exists to preserve exactly that behaviour: encoded paths, plain paths and a path with a literal percent are accepted; protocol-relative, encoded protocol-relative, absolute off-site with and without encoding, both backslash variants, `javascript:`, and malformed percent-encoding are refused.

**One reviewer recommendation I initially argued against and then took.** Codex wanted the write sanitised too; I had said the read sinks are the boundary and a wrong filter in untrusted JavaScript would silently break "come back here". Both are right about something, so the check went in at the single point where a value out of the URL bar becomes a stored destination, and nowhere else — the other writers take their values from hrefs the server emitted. It is explicitly not the boundary; all three reads stay guarded.

**And I shipped my own multi-line `{# #}` bug a second time**, in the very commit fixing sink #3. The phrase-list test I had written did not catch it, because a phrase list only catches instances someone remembered to add. Replaced with a structural scan of every template for multi-line `{# ... #}`, which caught it immediately. I also cut the long comment the previous commit put in `registration_base.html` — it is served to every visitor of those pages, and spelling out an unfixed vulnerability in page source is the opposite of what I should be doing while the fix is undeployed.

The label is off and stays off until both reviewers are clean on this head.

rdhyee:
**[CC]** Two reviews disagreed about whether the JavaScript path was exploitable. I settled it by running the code rather than picking a reading — head `50021c5a`.

### The disagreement

One review said `window.location.replace(saved_next)` navigated to an unvalidated cookie and took the user off-site. The other said the value is still `auth_next`-encoded at that point, so it merely 404s.

**Both are right, for different provenance, and the security-relevant one is the first.** When the inline script writes the cookie *straight from the URL's `?next=`*, `auth_next` never touched it — one layer of encoding, one decode on read, and `location.replace` would get `//evil.example` intact. The "it 404s" reading only holds for values that came through `auth_next` and the hijax lift.

### How I settled it

I extracted the **actual script served** from `/accounts/superlogin/welcome/` and ran it in node against a stub DOM, walking the real exploit: lure to `/accounts/register/?next=//evil.example`, register, land on the welcome page.

| Case | Cookie after | Navigated to | Off-site? |
|---|---|---|---|
| lure page stores the cookie | *not stored* | — | no |
| welcome page, cookie pre-poisoned `//evil.example` | `//evil.example` | `/next/` | no |
| welcome page, cookie pre-poisoned `https://evil.example/phish` | `https://evil.example/phish` | `/next/` | no |
| control: legitimate `/work/1/` | `/work/1/` | `/next/` | no — still works |

Two independent layers: the write-side check refuses to store it at all, and even a cookie planted by other means only reaches `/next/`, where the server guard sends it to `/` (covered by the Django tests). The exploit is closed, demonstrated rather than argued.

### The `require_https` finding was real, and a genuine downgrade

`require_https=request.is_secure()` **never fires in production**: TLS terminates at the proxy and `SECURE_PROXY_SSL_HEADER` is unset, so `is_secure()` is always `False`. `http://unglue.it/…` would have validated and bounced the user off TLS for that hop — while the docstring read as though the guard were active. A flag that looks live and is not is worse than no flag.

Fixed by removing the question rather than answering it: the target must now be a **relative same-site path** — no scheme, no netloc, no control characters. No absolute URL passes, http or https, ours or anyone's. That is also simply correct for this cookie, since every writer stores a path. Raw control characters are rejected too, because browsers strip tabs and newlines before resolving a URL, so a value that parses as a path in Python can resolve to something else in the browser.

The underlying infrastructure gap is filed as #1264 rather than fixed here.

### The overclaiming pattern, caught a second time

The test docstring called the remaining `?next=`-carrying Google links "a bounded URL space". They are not strictly bounded — `home.html` passes `request.GET.next` through, so `/?next=<anything>` mints a distinct Google URL. The true, narrower claim is that nothing on the site *links* to those templates with a varying `next`, so they are not a self-minting surface the way the header links were. Reworded.

This is the third time in this PR that a simplification made a claim bigger than the evidence. I am now treating it as the default failure mode rather than a one-off.

### Noted, not fixed

The welcome-page redirect costs one extra hop (`/next/`, then the destination) in exchange for server-side validation, and there is no `<noscript>` fallback there — the same JavaScript-off tradeoff as the header links, in a second place. Both are in the PR body.

rdhyee:
**[CC]** `/code-review` on head `50021c5a`: **no high-severity bug.** It traced every writer of the `next` cookie against the two-`unquote` read and confirmed the encoding contract holds for each, that off-site, protocol-relative, backslash, double-encoded, absolute-same-host and control-character variants are all rejected, and — usefully — that the new tests are **not vacuous**: `{% url 'next' %}` resolves, both `/accounts/*/welcome/` routes exist, and `django_registration/registration_complete.html` really does extend the patched template.

Four lower-severity findings. Two were real and shared one cause; both are fixed in `34785757`.

### `/next/` was a valid destination for the cookie that `/next/` reads

**Self-redirect.** A cookie holding `/next/` produced `Location: /next/`. I verified that before fixing rather than taking it on the report. It terminated only because the same response clears the cookie — so termination depended on the delete landing, and a cookie scoped to a narrower path (browsers send that one first, and `delete_cookie` on `/` will not remove it) would have looped.

**A real destination got clobbered.** This one is a genuine regression I introduced. The login page's fallback links now carry `?next=/next/`, and `registration_base.html` stores whatever `?next=` it sees. So a visitor who saved `/work/12345/` from the header Sign In, then passed through a bare login page and clicked "Need an account", had it overwritten with a pointer to nowhere. Also verified — `/accounts/register/?next=/next/` really is emitted.

Both closed by refusing `/next/` and anything under it, **on the server when reading and in the JavaScript when writing**. The two checks mirror each other deliberately.

Re-ran the exploit simulation against the changed script: still no off-site navigation in any case, the legitimate flow still works, and a saved `/work/12345/` now survives a page carrying `?next=/next/` instead of being overwritten.

### Noted, not fixed

- **`registration_complete.html` offers the "go back to the page where you were" link on cookie *presence*, not validity.** A user whose cookie holds something unusable is promised a return and lands on the home page. With the write-side check in place that case is now rare, and the link is harmless when it happens — `/next/` re-validates independently. Fixing it properly wants a context processor or template filter, which is more machinery than the case deserves.
- **No `<noscript>` path** for "sign in and come back here". This is the deliberate trade at the top of the PR description. The reviewer's sharpest point about it is worth repeating: the `hijax` lightbox on the same link makes the failure **silent rather than visible**, which is the part worth weighing.

### Where this leaves the PR

Both reviewers are now clean on high severity. I am not re-applying the `CC+Codex+LGTM` label myself — it came off at the team lead's instruction after the second sink was found, and it should go back on by their call, on this head, not mine.

rdhyee:
**[CC]** Final review round. `/code-review` on `d23d8965`: **no correctness bugs.** It probed the open-redirect work adversarially — protocol-relative, backslash, double- and triple-encoded, `javascript:`, scheme-with-empty-netloc, tab-injected — against both the server view and the JavaScript write-side check, and everything is rejected or lands same-origin. It confirmed all three read sinks closed and grepped for a fourth: none.

Two claims corrected in the description, both the same failure mode I have now hit repeatedly here.

### "One sign-in URL instead of two million" was too strong

The promotion rewrites the href **in the DOM**, so a crawler that renders JavaScript — Googlebot does — still sees one sign-in URL per page. And `rel="nofollow"` does not close that gap: it has been a hint rather than a directive since 2019.

What is true is narrower and still the point: the fan-out is eliminated **for crawlers that do not execute JavaScript**, which is the fleet that caused the 9/17 outages. The description now says it that way. This is the sentence that would have been quoted back at me.

### The no-JavaScript consequence, in plain terms

Now stated as an outcome rather than a mechanism: with JavaScript off or blocked — or if `sitewide1.js` simply fails to load — the header Sign In link stays the bare URL the server rendered, nothing tells the site where you came from, and after signing in you land on the home page and have to find your way back. On that link the `hijax` lightbox makes it fail quietly rather than visibly.

### One thing I checked instead of repeating

Review attributed the pre-existing test errors to `libraryauth/templatetags/puzzle.py` raising without a captcha key file, and expected them to pass in CI. **That is not the cause of any of them.** Traced individually:

| Test | Cause |
|---|---|
| `RhPageTests` ×3 | the regex looks for `name='csrfmiddlewaretoken'` in **single** quotes; Django renders double. Stale test — fails anywhere. |
| `TestLibraryAuth.test_registration` | `reverse('registration_complete')`; the name is `django_registration_complete`. Stale test — fails anywhere. |
| `AnonymousStripeCustomerBugTest` ×2 | Redis not running locally. Environment. |
| `TransactionTest.test_paymentmanager_charge` | no real Stripe key locally. Environment. |

Two are environment and would pass where Redis and a key are present; **two are genuinely broken tests that fail everywhere** and predate this branch. All seven are identical on a clean `origin/master` run against the same database. I would rather say that than "they pass in CI", which I cannot verify from here.

### Also noted

`frontend/templatetags/urldecode.py` is now dead — verified, the only template that loaded it was the one whose sink this PR removed. Deleting it is right and is a follow-up, but it should not sit long, since leaving it invites reintroduction of exactly the sink it enabled. The `?v=` cache-buster is manual; the template carries a note to bump it, and the test asserts only that *some* version marker is present, not that it tracks the file.

### Label

Applying `CC+Codex+LGTM`: both reviewers are clean on high severity and on correctness, on this head. **Not merging** — that is @rdhyee's, and the description's top blockquote is the thing to decide before it happens.

rdhyee:
**[CC]** Removing `CC+Codex+LGTM` again — I applied it prematurely and it is currently asserting something that is not true.

The label means *both reviewers clean on the final head*. Head is `34785757`. Codex last read `d23d8965`; `/code-review` last read `50021c5a`. **Neither has seen this head end-to-end**, and this head is not a cosmetic increment — it contains two real bugs found *after* the last full review, one of them a regression I introduced in the commit before it.

I applied the label on the earlier instruction to do so once the claims were fixed, and read that as still standing. It was the wrong read: the claims were only part of the condition, and I had just pushed two more commits.

Which is, precisely, the failure mode this PR has now hit four times — "the same URL on every crawlable page", "a bounded URL space", reporting the open redirect closed when one of three sinks was fixed, and now a label saying reviewed-clean on a head no reviewer has read. Every one of them was a claim that ran slightly ahead of the evidence, and every one was caught by someone else. A label is just a very short claim, and it gets the same standard.

A review of this head is running. It goes back on when that is true, and not by me.

rdhyee:
**[CC]** Rebased onto #1262 (`06d3fcc6`). New head `c7db15df`. The PR now shows 19 commits — 8 theirs, 11 mine — and will shrink back to 11 once #1262 merges first.

### Conflicts, and how they were resolved

Three, all small:

- **`feedback_link.py` docstring** — resolved toward keeping *both* facts, which is what I told them I would do: their point that `feedback_url` no longer needs its self-reference guard, because a constant URL cannot recurse; plus mine that `auth_next` still does, and that its value now rides in `data-next` rather than the href. The diff against their branch for that file is **docstring only** — I verified no code of theirs changed.
- **`frontend/tests.py`** — purely additive, both test classes kept.
- **`frontend/views/__init__.py`** — two import lines, both sides' imports merged (`add_never_cache_headers` and `urlsplit` are theirs, `url_has_allowed_host_and_scheme` and `urlparse` mine).

### The one test that had to change, and it did not

`test_feedback_login_chain_reaches_fixed_point` is the function we both edited. Git merged it cleanly: their `follow=True` on the first fetch, needed because their branch now 301s any non-POST request to `/feedback/` carrying a query string, sits alongside my change reading the sign-in value out of `data-next` rather than the href. Different lines, both required, neither weakened.

### Verified

- **112 tests**, with the same 3 failures and 4 errors as a clean `origin/master` baseline — the pre-existing set whose individual causes are tabled earlier in this thread.
- **53 tests** across both agents' classes specifically (`FeedbackSelfLinkTests`, `FeedbackUrlSpaceTests`, `SignInUrlSpaceTests`, `NextCookieRedirectTests`, `WelcomePageRedirectTests`, `GiftLoginNextTests`): all pass.
- **Their work is intact** — checked rather than assumed: `referrerpolicy` still on the feedback links, the `Referer`-reading logic still in `frontend/views`.
- **My diff on top of their head is sign-in only**, twelve files, nothing of theirs touched beyond that one docstring.

The force-push used `--force-with-lease` against the verified prior head, and I kept a local branch at the pre-rebase commit in case anything needs recovering.

Label deliberately **not** re-applied. A review of the previous head is still out, and a rebase changes the head, so whatever it says will be about a commit that no longer exists. That call belongs to the team lead, once, on this head.

rdhyee:
**[CC]** Both findings fixed on top of the rebase. Head **`fdae0d9a`**.

### The space bug — my guard broke pages that worked before this PR

`_is_safe_redirect_target` rejected any character `<= ' '`, which includes the **space**, on the fully double-decoded value. This site has free-text path routes — `/free/<path>/` for keyword facets, `/bypub/all/<pubname>` for publisher names — so `/bypub/all/Oxford University Press` and `/free/kw.science fiction/` are real pages a real user can be sitting on.

`auth_next` percent-encodes the space, `jquery.cookie` encodes it again, the view's two unquotes decode it back to a literal, and my guard then threw the destination away and dropped the visitor on the home page after signing in. **That worked before this PR touched any of it.**

Verified before fixing rather than taking it on the report:

| Path | old guard | new guard | `Location` emitted |
|---|---|---|---|
| `/bypub/all/Oxford University Press` | rejected | accepted | `/bypub/all/Oxford%20University%20Press` |
| `/free/kw.science fiction/` | rejected | accepted | `/free/kw.science%20fiction/` |
| `/work/1/` | accepted | accepted | `/work/1/` |

`HttpResponseRedirect` runs the target through `iri_to_uri`, which re-encodes the space on the way out — exactly what already happened. Narrowed to `ch < ' '` plus `\x7f`, so tabs and newlines are still refused; those are the ones browsers strip before resolving a URL, which is the actual risk the check exists for. Re-checked every hostile case: `//evil.example`, `https://evil.example`, and the tab and newline variants are all still rejected. Both real paths are now regression tests, and I falsified them — reverting the one character makes them fail.

**Worth naming plainly.** The write-side guard went in on Codex's advice, to limit exposure from future forgotten consumers, over my objection that a wrong filter on untrusted input silently breaks real users. Both positions turned out right, and the synthesis is the useful part: the check belongs there, *and* it needed testing against legitimate input. I did exactly that for the JavaScript twin — ran it against encoded paths, plain paths, a literal percent — and then didn't for its server-side counterpart. The bug lived in the half I skipped.

### The hijax writer was missing the `/next/` check

Same family. `registration_base.html` got the self-reference guard; the hijax cookie writer in `sitewide1.js` did not. A visitor on `/work/1/` clicks Sign In (cookie = `/work/1/`), follows "Need an account" to a page whose links carry `?next=/next/`, clicks Sign In again — cookie overwritten with a pointer to nowhere. The read guard makes that **safe, not harmless**: the destination is gone. Added, sharing a `decodeNextSafely` helper that tests a value without throwing on a malformed one and without altering what gets stored.

### Throwaways, both taken

`frontend/templatetags/urldecode.py` deleted — confirmed dead across every file type, and leaving it invites reintroducing the sink it enabled. Unused `dj_settings` import dropped.

### State

114 tests, same 3 failures and 4 errors as the clean baseline. Rebased on #1262 at `06d3fcc6`. Label deliberately not re-applied — that call belongs to the team lead, once, on this head.

rdhyee:
**[CC]** Label applied by the orchestrating session on head `fdae0d9a`, after verifying the code on GitHub rather than relying on reports.

What it rests on: Codex across multiple rounds at low effort; two `/code-review` passes, the second of which found the space-rejection regression (`/bypub/all/Oxford University Press` and `/free/kw.science fiction/` would have sent signed-in visitors to the home page — behaviour that works on master today) and the missing `/next/` guard in the hijax cookie writer; and the author's own node-executed walk of the real exploit against the served script, including a control case proving a legitimate destination still works.

What it does **not** cover: nothing was run against any server; no real browser or Google OAuth round trip was exercised; the three closed redirect sinks were verified by test and by grep for a fourth, not by pen-testing production. Two of the seven pre-existing test failures on this branch are stale tests that would fail in CI too (a CSRF regex expecting single quotes, and `reverse('registration_complete')` where the route is `django_registration_complete`) — neither is caused by this PR.

Open for @rdhyee and @eshellman, stated in the PR body: with JavaScript off, "sign in and come back here" lands you on the home page instead. And the fan-out is eliminated for crawlers that do not execute JavaScript — the fleet behind the 2026-09-17 outages — not for all crawlers.

eshellman:
CC i pulling your chain methinks

